PR #10, opened 2021-10-06
adds automatic pypi publishing

dowusu-antwi:
Adds yaml and shell script files to automate PyPI publishing steps in SupertechLabs/SuperstaQ#715.


vtomole (approved):
LGTM

PR #12, opened 2021-10-06
Automate pypi publish

dowusu-antwi:
Adds run privileges to pypi publish shell script.


vtomole (changes requested):
Why is this showing new files? We already added these files.

vtomole (approved):
LGTM

PR #27, opened 2021-12-30
Adds client method to set IBMQ token

dowusu-antwi:
Adding functionality for using new server endpoint, to set IBMQ token (closely following `aqt_upload_configs`, which added similar functionality).

This additional functionality is added to the `SuperstaQClient` object referenced by `css` and `qss`, and so is needed to set the IBMQ token for those repositories.


singular-value (approved):
lgtm

singular-value:
Might as well increment _version.py while you're at it, in advance of deploying a new applications-superstaq version?

richrines1 (commented):
should there be a corresponding function in user_config.py? (as there is for `aqt_upload_configs()`)

dowusu-antwi:
> should there be a corresponding function in user_config.py? (as there is for `aqt_upload_configs()`)

I'm not sure -- it looks like none of the `user_config` functions are making changes. Both functions are reading values and returning them to the user (?) while a `set_ibmq_token` method would actually change user information. Perhaps this is what we want in `user_config`.

richrines1:
> I'm not sure -- it looks like none of the `user_config` functions are making changes. Both functions are reading values and returning them to the user (?) while a `set_ibmq_token` method would actually change user information. Perhaps this is what we want in `user_config`.

I think the idea would be for `user_config.set_ibmq_token(token: str)` to be the user-facing interface - so take a single string as input, and call away to `self._client.set_ibmq_token({"ibmq_token": token})`. This is roughly what `user_config.aqt_upload_configs` does; it takes filenames and then calls away to `self._client.aqt_upload_configs({"pulses": pulse_data, "variables": variable_data})` which uploads the data to the server

richrines1:
a few other questions:
- would it makes sense to make `token` optional in the `UserConfig` function and check for `.qiskitrc` if it isn't provided?
- there's a bit of a naming inconsistency between `aqt_upload_configs` and `set_ibmq_token` - should it be `ibmq_set_token` or `ibmq_upload_token`? (not sure how I feel about this)
- beyond the scope of this PR, but I'm noticing many functions in `superstaq_client` are identical up to the endpoint name. Would it make sense to merge these into something like `superstaq_client.post_request(endpoint, json_dict)`, so e.g. `UserConfig.set_ibmq_token` would call `self._client.post_request("ibmq_token", {...})`

singular-value:
> there's a bit of a naming inconsistency between aqt_upload_configs and set_ibmq_token - should it be ibmq_set_token or ibmq_upload_token? (not sure how I feel about this)

IMO the upload vs set verb is appropriate, because aqt_upload_configs uploads the contents of two whole files.
But yeah, let's change the order to `ibmq_set_token` for consistency

vtomole:
> Would it make sense to merge

Yup!

dowusu-antwi:
> Would it make sense to merge
 
Just made an issue for this, #28. We might want to complete this PR sooner so we can make sure to deploy (given the incrementing of `_version.py`), but whether we make those changes here or in a separate PR we can link it to #28.

richrines1:
^ cool sounds good. final question: currently it seems like we mostly use `ibmq_*` elsewhere for IBM-related methods; should these also be `ibmq_set_token` instead of `ibm_set_token`?

dowusu-antwi:
Yeah good idea. In fact, I was thinking we should probably use `ibmq` across the board, since IBM does a lot of things beyond quantum, lol.

richrines1:
> Yeah good idea. In fact, I was thinking we should probably use `ibmq` across the board, since IBM does a lot of things beyond quantum, lol.

good point, agreed.

richrines1 (approved):
lgtm

Issue #28, opened 2021-12-30
Merge identical `superstaq_client` methods


dowusu-antwi:
From Rich's comments on PR #27: many functions in superstaq_client are identical up to the endpoint name. Avoid copy-paste code by merging these into `superstaq_client.post_request(endpoint, json_dict)`.

vtomole:
I'll take this.

PR #30, opened 2022-01-13
`/cq_compile endpoint` and consoliate POST/GET requests.

vtomole:
Fixes: https://github.com/SupertechLabs/applications-superstaq/issues/28


applications_superstaq/superstaq_client.py
```diff
@@ -160,95 +175,39 @@ def get_job(self, job_id: str) -> dict:
...
-            return requests.post(
-                f"{self.url}/qscout_compile",
-                headers=self.headers,
-                json=json_dict,
-                verify=self.verify_https,
-            )
-
-        return self._make_request(request).json()
+    def cq_compile(self, json_dict: Dict[str, Union[str, List[str]]]) -> dict:
+        """Makes a POST request to SuperstaQ API to compile a list of circuits for ColdQuanta."""
```
singular-value, line 201:
```suggestion
        """Makes a POST request to SuperstaQ API to compile a list of circuits for CQ."""
```

richrines1:
are `get_request`/`post_request` always wrapped in `_make_request`? if so, would it make sense to just call `_make_request` from those functions?

vtomole:
@richrines1 Done.

applications_superstaq/superstaq_client.py
```diff
@@ -98,25 +98,46 @@ def __init__(
...
         target: Optional[str] = None,
-        name: Optional[str] = None,
         ibmq_token: Optional[str] = None,
         ibmq_pulse: Optional[bool] = None,
     ) -> dict:
         """Create a job.
 
         Args:
+            ibmq_pulse: Specify whether to run the job on the pulse level of the device
+            ibmq_token: For running on IBM devices
```
singular-value, line 134: nit: move these down, to match the order (ibmq_token and then ibmq_pulse last)

applications_superstaq/superstaq_client.py
```diff
@@ -277,81 +234,28 @@ def submit_qubo(self, qubo: qv.QUBO, target: str, repetitions: int = 1000) -> di
...
-
-        def request() -> requests.Response:
-            return requests.post(
-                f"{self.url}/warehouse",
-                headers=self.headers,
-                json=json_dict,
-            )
-
-        return self._make_request(request).json()
+        return self.post_request("/warehouse", json_dict)
```
singular-value, line 254: I just noticed that we don't have a `verify` flag in the previous warehouse post request, but that's probably a bug/inconsistency (that this PR fixes)

applications_superstaq/superstaq_client.py
richrines1: why is `ibmq_token` being passed here at all? I thought this was handled via `ibmq_set_token()` now

applications_superstaq/superstaq_client.py
vtomole: Yup. An accidental bug fix.

applications_superstaq/superstaq_client.py
vtomole: This passed through https://github.com/SupertechLabs/applications-superstaq/pull/27. Just removed.

perlinm:
Hmm... `check/mypy` runs fine on my local machine, but on the server it seems to complain about how we initialize arrays on line [9](https://github.com/SupertechLabs/applications-superstaq/blob/952f0b7143846047572ef0f8a6db0a11a37d9244/applications_superstaq/qubo_test.py#L9) of `qubo_test.py` and lines [52](https://github.com/SupertechLabs/applications-superstaq/blob/952f0b7143846047572ef0f8a6db0a11a37d9244/applications_superstaq/finance_test.py#L52) and [64](https://github.com/SupertechLabs/applications-superstaq/blob/cq_compile_endpoint/applications_superstaq/finance_test.py#L64) of `finance_test.py`.  Any ideas for what's going on?

vtomole:
We inherit our numpy version from qubovert which [doesn't pin a version of numpy ](https://github.com/jtiosue/qubovert/blob/d5ea46349d2a058954fb2cb06f559c0d3fb382c5/requirements.txt#L1). The most recent release of numpy added typing that breaks our code. 

We could either fix the type error or pin the numpy version.

perlinm:
In that case imo pinning the version of `numpy` is hacky, and we should figure out how to make `numpy`/`mypy` happy here.

perlinm:
Ah looks like `superstaq` already requires `requires numpy==1.21.*`, so I guess we could do the same here for compatibility.  Not sure I like that though.

perlinm:
I think changing `np.rec.array` to `np.core.records.fromrecords` fixes the problem.

vtomole:
Very nice. Thanks!

applications_superstaq/superstaq_client.py
dowusu-antwi: Yep, I missed that one in #27 -- thanks!

richrines1:
> It looks like changing `np.rec.array` to `np.core.records.fromrecords` fixes the problem.

nit: use `np.rec.fromrecords` (`np.core` is [considered private](https://github.com/numpy/numpy/blob/c852e1397bf3c0b24faaa3c66cc50b1f2eba1ebd/numpy/core/__init__.py#L4))

vtomole:
I think we should pin the numpy version anyway so we have a say on when upstream package updates breaks our checks.

perlinm:
The checks only matter for development (despite `mypy`'s complaints, the actual `pytest` ran without a problem), and keeping `numpy` un-pinned keeps us up-to-date / on our toes.  As it concerns actually running (as opposed to testing) code -- in some sense it's on `numpy` to not break backwards compatibility.

Just my two cents though.  If you still want to pin `numpy` that's fine by me (and it's consistent with what `superstaq` already does).

vtomole:
Okay sounds good. Will leave as-is.

applications_superstaq/superstaq_client.py
```diff
@@ -116,7 +135,7 @@ def create_job(
                 from the returned job.
             target: If supplied the target to run on. Supports one of `qpu` or `simulator`. If not
                 set, uses `default_target`.
-            name: An optional name of the job. Different than the `job_id` of the job.
+            ibmq_pulse: Specify whether to run the job on the pulse level of the device
```
richrines1, line 138:
```suggestion
            ibmq_pulse: Specify whether to run the job on the pulse level of the device
```
```suggestion
            ibmq_pulse: Specify whether to run the job on the pulse level of the device
```

richrines1, line 138: imo it isn't obvious what running "on the pulse level of the device" means - would something like "using SuperstaQ pulse-level optimizations" make sense?

vtomole: That is better. Done.

richrines1:
lgtm, leaving for @perlinm

perlinm:
LGTM

perlinm (approved):

perlinm:
I presume we need to deploy between these [cirq-superstaq](https://github.com/SupertechLabs/cirq-superstaq/pull/154) and [qiskit-superstaq](https://github.com/SupertechLabs/qiskit-superstaq/pull/84) PRs will pass?

vtomole:
Yes, I'll deploy now.

applications_superstaq/superstaq_client.py
```diff
@@ -116,7 +135,7 @@ def create_job(
                 from the returned job.
             target: If supplied the target to run on. Supports one of `qpu` or `simulator`. If not
                 set, uses `default_target`.
-            name: An optional name of the job. Different than the `job_id` of the job.
```
dowusu-antwi, line 119: (Removing this parameter required changes to `cirq-superstaq`, see https://github.com/SupertechLabs/cirq-superstaq/pull/159.)

PR #33, opened 2022-01-25
Added message attribute to `SuperstaQException`

dowusu-antwi:
This is so we can access the exception message independently of the status code (e.g., useful for displaying top-layer error messages with minimal clutter).

### DIFF
--- a/applications_superstaq/__init__.py
+++ b/applications_superstaq/__init__.py
@@ -1,7 +1,6 @@
 from applications_superstaq._init_vars import API_URL, API_VERSION
 from applications_superstaq.superstaq_exceptions import (
     SuperstaQException,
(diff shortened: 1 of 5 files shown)


dowusu-antwi:
(Offline discussion) We've been returning status codes twice (see #34). Currently fixing.

singular-value (approved):
lgtm

dowusu-antwi:
Note: from discussion, we found there are two cases for the server response object `requests.Response`: either it contains a `JSON` with the message as a key-value pair, or it doesn't and the message is just `response.text`.

I also added a test for the case in which the `JSON` contains the message.

vtomole (approved):
LGTM

PR #43, opened 2022-04-06
`aqt_get_configs` endpoint call

vtomole:


applications_superstaq/user_config_test.py
```diff
@@ -50,3 +50,23 @@ def test_service_aqt_upload_configs(mock_aqt_compile: mock.MagicMock) -> None:
...
+    return_value={"pulses": "Hello", "variables": "World"},
+)
+def test_service_aqt_get_configs(mock_aqt_compile: mock.MagicMock) -> None:
+
+    client = applications_superstaq.superstaq_client._SuperstaQClient(
+        remote_host="http://example.com", api_key="key", client_name="applications_superstaq"
+    )
+    service = applications_superstaq.user_config.UserConfig(client)
+
+    service.aqt_get_configs("/tmp/Pulses.yaml", "/tmp/Variables.yaml")
```
richrines1, line 66: this isn't really portable - we should probably be using something like [tempfile](https://docs.python.org/3/library/tempfile.html) instead of hardcoding the file path (also should be changed in `test_service_aqt_upload_configs()`)

vtomole: > this isn't really portable  

How so? `/tmp` is guaranteed to exist in every POSIX system.

richrines1: shouldn't clients work on windows as well?

vtomole: Did we ever guarantee this? We should run a Windows Github actions if so. I wouldn't be surprised if our clients don't work on Windows right now.

vtomole: I'm down to only look into Windows support if someone asks but not before.

richrines1: i guess other than these lines i don't see any obvious reason why the clients wouldn't work as they are, but yeah we'd definitely need to test it. i'd be a little surprised if no one has run a demo/example notebook on windows already

definitely getting outside the scope of this PR though. meanwhile, if we're going to make it explicitly posix-only we should maybe just update setup.py to reflect that?

applications_superstaq/user_config.py
```diff
@@ -52,3 +52,20 @@ def aqt_upload_configs(self, pulses_file_path: str, variables_file_path: str) ->
...
+        """Writes AQT configs from the AQT system onto the given file paths.
+
+        Args:
+            pulses_file_path: where to write the pulse configurations
+            variables_file_path: where to write the variables configurations
+        Returns:
+            None
+        """
+        config_dict = self._client.aqt_get_configs()
+        with open(pulses_file_path, "w") as text_file:
```
richrines1, line 66: do we want to prevent this from overwriting an existing file?

applications_superstaq/user_config.py
```diff
@@ -52,3 +52,20 @@ def aqt_upload_configs(self, pulses_file_path: str, variables_file_path: str) ->
...
+            text_file.write(config_dict["pulses"])
+
+        with open(variables_file_path, "w") as text_file:
+            text_file.write(config_dict["variables"])
+        return None
```
richrines1, line 71:
```suggestion
```

applications_superstaq/user_config.py
vtomole: https://github.com/SupertechLabs/applications-superstaq/issues/44

applications_superstaq/user_config_test.py
vtomole: https://github.com/SupertechLabs/applications-superstaq/issues/44

applications_superstaq/user_config.py
vtomole: Thanks. Don't remember why i added this. I assumed mypy was strict about returning the types but that might be a rule we added to our internal stuff.

applications_superstaq/user_config.py
vtomole: Done.

applications_superstaq/user_config.py
```diff
@@ -52,3 +53,38 @@ def aqt_upload_configs(self, pulses_file_path: str, variables_file_path: str) ->
...
+            variables_file_path: where to write the variables configurations
+        Returns:
+            None
+        """
+        pulses_file_exists = os.path.exists(pulses_file_path)
+        variables_file_exists = os.path.exists(variables_file_path)
+
+        if pulses_file_exists and not variables_file_exists:
+            raise ValueError(
+                f"{pulses_file_path} exists as a pulse file. "
```
richrines1, line 71: nit
```suggestion
                f"{pulses_file_path} exists. "
```
(e.g. if i write `aqt_get_configs("Documents/my-grocery-list", ...)` it might exist but not "as a pulse file")

(ditto below)

applications_superstaq/user_config.py
```diff
@@ -52,3 +53,38 @@ def aqt_upload_configs(self, pulses_file_path: str, variables_file_path: str) ->
...
+                f"Please try a different filename to write to"
+            )
+        elif not pulses_file_exists and variables_file_exists:
+            raise ValueError(
+                f"{variables_file_path} exists as a variables file. "
+                f"Please try a different filename to write to"
+            )
+        elif pulses_file_exists and variables_file_exists:
```
richrines1, line 79: nit: if this check was first you could skip the `not *_exists` in the other two `if` statements

applications_superstaq/user_config.py
```diff
@@ -52,3 +53,38 @@ def aqt_upload_configs(self, pulses_file_path: str, variables_file_path: str) ->
...
+            raise ValueError(
+                f"{pulses_file_path} and {variables_file_path} "
+                f"exists as pulses and variable files. Please try different filenames to write to"
```
richrines1, line 82: nit
```suggestion
                f"exist as pulses and variable files. Please try different filenames to write to"
```

applications_superstaq/user_config.py
vtomole: Modified.

applications_superstaq/user_config.py
vtomole: Done.

richrines1:
one more thought: what do you think about calling this `aqt_save_configs`, and having `aqt_get_configs()` just return the two configs as dictionaries?

applications_superstaq/user_config_test.py
```diff
@@ -12,7 +16,7 @@ def test_service_get_balance() -> None:
     mock_client.get_balance.return_value = {"balance": 12345.6789}
     service._client = mock_client
 
-    assert service.get_balance() == "$12,345.68"
+    assert not service.get_balance() == "$12,345.69"
```
richrines1, line 19: ?

vtomole: Whoops reverted.

vtomole:
> having aqt_get_configs() just return the two configs as dictionaries?

It would look like 
```python
 def aqt_get_configs(self) -> Dict:
        return self._client.aqt_get_configs()
```

Which doesn't seem like it's simplifying anything. We can rename our `aqt_get_configs` to `aqt_save_configs`.

richrines1:
> Which doesn't seem like it's simplifying anything. 

true. it would also make it accessible via `css.Service.get_aqt_configs()`/etc., but i agree this maybe isn't super helpful

> We can rename our `aqt_get_configs` to `aqt_save_configs`

i think this might be a good idea. it's more descriptive of what it does, and differentiates it from `_client.aqt_get_configs()` which doesn't do the same thing

vtomole:
> it would also make it accessible via css.Service.get_aqt_configs()/etc., but i agree this maybe isn't super helpful

Maybe some just needs to see that dict instead saving it to a file to see it. I've added it.

vtomole:
Renamed `aqt_get_configs` to `aqt_save_configs`.

richrines1 (approved):
nice, lgtm

PR #45, opened 2022-04-07
replace `check/format` with `check/format.py`

perlinm:
see [issue](https://github.com/SupertechLabs/qiskit-superstaq/issues/140)

~also removing check/pytest github action~ (reverted; this should be dealt with separately after [issue](https://github.com/SupertechLabs/SuperstaQ/issues/1143)/[PR](https://github.com/SupertechLabs/SuperstaQ/pull/1181))


check/format.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
vtomole, line 1: We don't package the `check` code in pypi. Let's move this into `applications_superstaq`. That way, downstream repos will do

```python
import applications_superstaq

applications_superstaq.format()
```

perlinm: I'm super confused.  The check scripts are command-line utilities.

perlinm:
Closing this PR.  We will instead move all check scripts from `SuperstaQ` to here at a later date.

PR #57, opened 2022-05-12
added windows pytest to applications-superstaq

PalashGoiporia:


vtomole:
@PalashGoiporia Please add the macos checks to the other repos as well.

vtomole:
Note to @richrines1. We expected this check to fail due to https://github.com/SupertechLabs/applications-superstaq/pull/43#discussion_r844390052 but it passed.

vtomole:
@PalashGoiporia @richrines1 Oh, that's because the tests are actually not running at all. See https://github.com/SupertechLabs/applications-superstaq/runs/6409726654?check_suite_focus=true

PR #70, opened 2022-06-16
migrating check scripts from SuperstaQ

### LINKED ISSUE #69: Migrate check scripts from SuperstaQ

perlinm:
Complete list of changes as of 2022.06.28:
- removed all old check scripts
- `cp ./Superstaq/check/*.py ./applications-superstaq/applications_superstaq/check/`
- removed SuperstaQ-specific environment variable definitions from `pytest_.py`
- exclude `dev_tools/*` by default from integration tests in `pytest_.py`
- `build_docs.py` now only tries to build docs if `[repo_root]/docs` exists (and is a directory)
- added `applications_superstaq/checks/__init__.py`
- updated how check scripts import each other
- updated workflows, which includes adding `pip install .` (see discussions from [here](https://github.com/SupertechLabs/applications-superstaq/pull/70#issuecomment-1168079610) onward)
- make `applications-superstaq/check` a symbolic link to `applications-superstaq/applications_superstaq/check`
- added `pytest-socket~=0.4.1` to `dev-requirements.txt`, updated other package versions to match requirements in SuperstaQ
- fixed requirements files with `check/requirements.py --apply`

fixes #69


vtomole:
We want these checks to be importable in the packages that `import applications_superstaq`. Let's move all these files to a `applications-superstaq/applications_superstaq/check` directory and have; for example `applications-superstaq/check/pylint_.py` be like 

```python
import applications_superstaq

if __name__ == "__main__":
    exit(applications_superstaq.pylint_.run(*sys.argv[1:]))
```

.github/workflows/ci.yml
```diff
@@ -132,7 +134,7 @@ jobs:
           if (Test-Path "dev-requirements.txt") {pip install -r dev-requirements.txt}
       - name: Pytest-win check
         run: |
-          check/pytest
+          check/pytest_.py
         shell: bash
```
perlinm, line 138: Pretty sure this line is no longer correct, but I'm not sure what should go there instead.  `python`?

.github/workflows/ci.yml
```diff
@@ -26,7 +26,8 @@ jobs:
           if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
       - name: Format
         run: |
-          check/format
+          check/format_.py
```
vtomole, line 29: Thoughts on moving all these to a `check.py` so we can call `python check.py --format`?

vtomole: We might not even need a check directory if we go this route.

perlinm: I use tab completion a lot, so I still want scripts in a check directory: `./ch<TAB>fo<TAB>` --> `./check/format_.py`

.github/workflows/ci.yml
perlinm: @vtomole?

vtomole: Why is it no longer correct?

vtomole: Discussed IRL: this is us telling Windows to use bash to run this check.

perlinm:
@vtomole do you mean to do that in this PR?

vtomole:
> are you saying we should do that in this PR?

Yes, most of the code on here has already gone through review. Given that, the PR won't be too big when we add tests cause we're mostly going to the focused on the implementation of the tests while reviewing.

perlinm:
Tagging @pgoiporia to make sure we incorporate changes recently made to SuperstaQ checks.

.github/workflows/ci.yml
```diff
@@ -26,7 +26,8 @@ jobs:
           if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
       - name: Format
         run: |
-          check/format
+          applications_superstaq/check/format_.py
```
vtomole, line 29: With the symlinks, why can't we have `check/format_.py` here for consistency with downstream repos?

perlinm: done

perlinm:
Observation: `[bash_test_scripts.py`](https://github.com/quantumlib/Cirq/blob/master/dev_tools/bash_scripts_test.py) does **not** cover all of the test files.  It only tests a limited amount of functionality.  Evidence: the only files checked by `bash_scripts_test.py` are:
- `check/check/pytest-changed-files`
- `check/pytest-and-incremental-coverage`
- `check/format-incremental`
- `check/pylint-changed-files`
There are [a lot of other check scripts](https://github.com/quantumlib/Cirq/tree/master/check) that are not checked.  I think trying to check and cover all of our checks might be too ambitious... and certainly more ambitious than what Cirq does.

vtomole:
Okay. Given that we are not good at [testing system code](https://github.com/SupertechLabs/SuperstaQ/issues/1392) anyway, I'm fine with us not checking checks and monitoring how they do. If they end up being really buggy (We find ourselves constantly fixing issues with them) then we'll just write tests for the regressions.

dev-requirements.txt
```diff
@@ -1,7 +1,9 @@
...
-mypy==0.910
-pylint==2.6.0
-pytest==6.1.2
+flake8==3.8.4
+mypy==0.961
+nbmake~=1.3.0
+pylint>=2.13.0
 pytest-cov==2.11.1
+pytest-socket~=0.4.1
+pytest==6.1.2
```
vtomole, line 9:
```suggestion
pytest~=6.1.2
```

dev-requirements.txt
```diff
@@ -1,7 +1,9 @@
 black[jupyter]~=22.3.0
-flake8==3.8.4
 flake8-import-order==0.18.1
-mypy==0.910
-pylint==2.6.0
-pytest==6.1.2
+flake8==3.8.4
```
vtomole, line 3:
```suggestion
flake8~=3.8.4
```

dev-requirements.txt
```diff
@@ -1,7 +1,9 @@
 black[jupyter]~=22.3.0
-flake8==3.8.4
 flake8-import-order==0.18.1
```
vtomole, line 2:
```suggestion
flake8-import-order~=0.18.1
```

vtomole (changes requested):
A few loosening of requirements then LGTM.

Before we merge, have we tested that this works on downstream repos?

- Install this version of app_ss locally
-  Import these checks in either qss,css, ss. Whatever you like; preferably all of them.
-  Confirm that these checks do the exact same thing as the checks we're about to yank following the merging of this.

dev-requirements.txt
perlinm: done

dev-requirements.txt
perlinm: done

dev-requirements.txt
perlinm: done

perlinm:
@vtomole good catch at testing it out.  This caught one minor issue and raised another one I'm less sure about how to fix: if we say, `import check_utils` in `format_.py`, that works fine within `apps_ss`, but then when I `import applications_superstaq.check` in `ss` I get `ModuleNotFoundError: No module named 'check_utils'`.

This can be fixed locally with `from applications_superstaq.check import check_utils` in `format_.py` (which is similar to, say, [`from applications_superstaq import superstaq_client` in `apps_ss/finance.py`](https://github.com/SupertechLabs/applications-superstaq/blob/c74cf945d2babe473374b2f99b55c7c54133e797/applications_superstaq/finance.py#L8)), but then the github checks don't work because of [this issue](https://github.com/SupertechLabs/applications-superstaq/pull/70#issuecomment-1158105594).  Possible resolution: install `apps_ss` on the server before running checks?  There could be a better way to fix this though.

richrines1:
> This can be fixed locally with `from applications_superstaq.check import check_utils` in `format_.py` (which is similar to, say, [`from applications_superstaq import superstaq_client` in `apps_ss/finance.py`](https://github.com/SupertechLabs/applications-superstaq/blob/c74cf945d2babe473374b2f99b55c7c54133e797/applications_superstaq/finance.py#L8)), but then the github checks don't work because of [this issue](https://github.com/SupertechLabs/applications-superstaq/pull/70#issuecomment-1158105594). Possible resolution: install `apps_ss` on the server before running checks? There could be a better way to fix this though.

not positive but i think you just need to use [relative imports](https://docs.python.org/3/reference/import.html#package-relative-imports), i.e. something like `from . import check_utils`

perlinm:
Unfortunately that gives me `ImportError: attempted relative import with no known parent package` :confused:

perlinm:
Alright so this worked:
```python3
if __package__ is None:
    import check_utils
else:
    from applications_superstaq.check import check_utils
```

@vtomole @richrines1 is that acceptable?

vtomole:
>  install apps_ss on the server before running checks?

We are not already doing this?

perlinm:
> We are not already doing this?

Nope, we [`pip install -r requirements.txt`](https://github.com/SupertechLabs/applications-superstaq/blob/c74cf945d2babe473374b2f99b55c7c54133e797/.github/workflows/ci.yml#L25).  That might(?) be reasonable, as I think I saw a similar thing in GitHub documentation somewhere yesterday, although I can't find the link now.

perlinm:
I don't see how that solves the problem.  The github workflows don't install `applications_superstaq` so they can't run `applications_superstaq.check.format_.run`.

richrines1:
^ ohhh my bad, i misunderstood [this](https://github.com/SupertechLabs/applications-superstaq/pull/70#issuecomment-1168895923)

this seems like a bug to me - it seems like we should definitely be installing the package instead of just reading the requirements. does doing so fix everything?

perlinm:
I think so, let me check.

vtomole:
^^
+1

vtomole (commented):
Does everything work now? Will approve if so.

perlinm:
Yep, everything works :slightly_smiling_face:

vtomole (approved):

PR #72, opened 2022-06-22
Add Resource Estimate request code 

paaige:


vtomole (commented):
We usually implement functionality in SuperstaQ first before building the API, so I'll review this once we merge the resource estimation code into SuperstaQ.

paaige:
> We usually implement functionality in SuperstaQ first before building the API, so I'll review this once we merge the resource estimation code into SuperstaQ.

That might be an issue because there's a dependency in superstaq (ResourceEstimate) that relies on the class being defined in applications_superstaq

vtomole:
Define it in SuperstaQ as well. We will replace it with this one when it's all set and done.

singular-value:
> Define it in SuperstaQ as well. We will replace it with this one when it's all set and done.

@vtomole our plan had been:
- get all four PRs ready for simultaneous review, except that qss/css/ss won't pass checks on GitHub because ResourceEstimate is not yet on applications_ss pypi. Paige can validate that checks pass on her local machine instead.
- deploy applications_ss pypi
- deploy qss/css/ss now that tests pass

Thoughts?

vtomole:
Sounds good.

vtomole (approved):
LGTM

singular-value:
Please don't merge in yet though. We may want to beef up the `ResourceEstimate` class first.

vtomole:
I'm fine with getting the API up as an MVP and adding more features as needed in subsequent PRs, but will leave on yawls court.

richrines1:
obviously we'll be expanding `ResourceEstimate` down the road but we might want to add something like a "num_global_gates` now --- the dominant gate in circuits compiled for CQ is the global_rgate, which won't be included in either "num_single_qubit_gates" or "num_two_qubit_gates"

alternatively, is there an easy way to break up gates by type? this would be ideal because otherwise for example `num_single_qubit_gates` doesn't tell us much on a system w/ both physical rxs and virtual rzs

singular-value:
@richrines1 agreed. However, we will deploy this right now so we can collect feedback and then iterate.

PR #75, opened 2022-06-29
make file argument optional in check scripts

### LINKED ISSUE #74: Incremental check runs on entire repo (bug)
If an incremental check (`./check/* -i`) finds no relevant changed files, it should check nothing instead of checking everything.  The problem comes down to lines such as:
```python
    if not files:
        files = check_utils.get_tracked_files(*default_files_to_check, exclude=exclude)
```

perlinm:
fixes #74


vtomole (approved):
LGTM

Issue #78, opened 2022-06-30
explicitly pass config files to `checks`


perlinm:
Basic idea is that if we make changes to `apps-ss/setup.cfg`, we want that change reflected in downstream repos.

perlinm:
Kicking the can on this one, as we're not sure whether we want to do it

perlinm:
Closing, will re-open if we decide we want it in the future

PR #79, opened 2022-06-30
update version after adding `[dev]` install option

perlinm:


perlinm:
Update: we might not want #78 after all, so we should just push the addition of a `[dev]` option for now

vtomole (approved):
LGTM. Don't forget to create a release once you've published.

Issue #84, opened 2022-07-01
`./check/pytest_.py -i` doesn't test anything if no `*_test.py` files have changed


perlinm:
The machinery to fix this issue exists in `./check/coverage_.py` (which works properly with `-i`), but this machinery is not used in `./check/pytest_.py`.

perlinm:
Note: we can tackle this at the same time as #80

Issue #66, opened 2022-06-13
`examples_test.py`


vtomole:
Similar to the notebooks we don't run our examples in the CI.

perlinm:
What do you mean by this?  I believe the new check scripts check everything in `examples` as well

vtomole:
Our `examples` directories would currently fail coverage checks if we turned them on.

perlinm:
Are you proposing to cover all examples?  I strongly disagree.  In practice, this just means I'll never put anything researchy I do into `examples`.

perlinm:
Actually: it looks like the alternate interpretation is what would occur if we simply removed `"examples/*"` from `default_exclude` [here](https://github.com/SupertechLabs/applications-superstaq/blob/fd170b1a4f1db0682795cabbec86682bb41f56c1/applications_superstaq/check/coverage_.py#L14).  That is, this may be a one-line fix.

vtomole:
> Are you proposing to cover all examples?

Sorry `coverage` is a bit loaded here. I mean we should actually run the examples in checks to make sure they execute. The same way` we do notebook_tests`.

vtomole:
Ah, idk how this will work for qiskit-runtime, but yeah we can selectively turn this off, the way we currently do notebooks we don't know how to execute.

perlinm:
Oh I see.  Unsure how to do that.  Two potential problems:
- `ss/examples/qiskit-runtime/*` stuff currently *does not run* because [conflicts](https://github.com/SupertechLabs/SuperstaQ/pull/1179#issuecomment-1176712709)
- Even if we fix the version conflicts above, `ss/examples/qiskit-runtime/train_kernel.py` and `plot_results.py` take a long time to run.  Seems wasteful to have integration tests re-run simulations all of the time.

It sounds like maybe I still don't understand what this issue is asking to do.

vtomole:
We can skip things that are unreasonable to run promptly like the qiskit runtime examples. What i had in mind when creating this issue is a check that makes sure https://github.com/SupertechLabs/cirq-superstaq/blob/main/examples/cirq_superstaq_example.py executes... but that also runs in the same problem; that is, that example submits circuits to IBM when we actually don't want to do that just for a test. Basically, we can't make progress on this until https://github.com/SupertechLabs/cirq-superstaq/issues/263 is done.

pgoiporia:
[...]

PR #92, opened 2022-07-08
Minor check fixes

### LINKED ISSUE #80: respect file arguments passed to `check/pytest_.py` in integration mode
it would be nice if `check/pytest_.py --integration <file>_test.py` just ran `<file>_test.py`

perlinm:
fixes #80 and #84 
also:
- make `check.requirements.run` use bash-style (as opposed to re-style) regexp for the `upstream_match` argument
- remove `examples/*.py` from the default excludes for `check/coverage_.py` (see #66)
- simplify how `check` scripts parse file/directory arguments (see discussions in the comments below)
- make `check/pylint_.py` incremental by default (with `-a/--all` options to check everything in the repo)

### DIFF
--- a/applications_superstaq/check/all_.py
+++ b/applications_superstaq/check/all_.py
@@ -1,9 +1,7 @@
 #!/usr/bin/env python3
 
-import argparse
 import sys
 import textwrap
-from typing import Optional
 
 from applications_superstaq.check import (
     build_docs,
@@ -17,13 +15,9 @@ from applications_superstaq.check import (
 )
 
 
-def run(
-    *args: str,
-    parser: Optional[argparse.ArgumentParser] = None,
-) -> int:
-    if not parser:
-        parser = check_utils.get_file_parser()
+def run(*args: str) -> int:
 
+    parser = check_utils.get_file_parser()
     parser.description = textwrap.dedent(
         """
         Runs all checks on the repository.
@@ -45,36 +39,26 @@ def run(
         dest="force_all",
         help="'Hard force' ~ continue past (i.e. do not exit after) all failing checks.",
     )
-    parser.add_argument(
-        "-i",
-        "--incremental",
-        dest="revisions",
-        nargs="*",
-        action="extend",
-        help="Run checks in incremental mode.",
-    )
-    parsed_args = parser.parse_intermixed_args(args)
-    files = parsed_args.files
-    revisions = parsed_args.revisions
 
-    default_mode = not files and revisions is None
+    parsed_args = parser.parse_intermixed_args(args)
+    args_to_pass = parsed_args.files
+    if parsed_args.revisions is not None:
+        args_to_pass += ["-i", *parsed_args.revisions]
+
+    default_mode = not parsed_args.files and parsed_args.revisions is None
     checks_failed = 0
 
     # run formatting checks
     exit_on_failure = not (parsed_args.force_formats or parsed_args.force_all)
-    checks_failed |= format_.run(*files, revisions=revisions, exit_on_failure=exit_on_failure)
-    checks_failed |= flake8_.run(*files, revisions=revisions, exit_on_failure=exit_on_failure)
-    checks_failed |= pylint_.run(*files, revisions=revisions, exit_on_failure=exit_on_failure)
+    checks_failed |= format_.run(*args_to_pass, exit_on_failure=exit_on_failure)
+    checks_failed |= flake8_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+    checks_failed |= pylint_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
 
     # run typing and coverage checks
     exit_on_failure = not parsed_args.force_all
-    # typing changes are likely to have side effects, so always run mypy on the entire repo
-    checks_failed |= mypy_.run(revisions=revisions, exit_on_failure=exit_on_failure)
+    checks_failed |= mypy_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
     checks_failed |= coverage_.run(
-        *files,
-        revisions=revisions,
-        exit_on_failure=exit_on_failure,
-        suppress_warnings=default_mode,
+        *args_to_pass, exit_on_failure=exit_on_failure, silent=default_mode
     )
 
     # check that all pip requirements files are in order
--- a/applications_superstaq/check/all_.py
+++ b/applications_superstaq/check/all_.py
@@ -64,12 +64,7 @@ def run(
     exit_on_failure = not (parsed_args.force_formats or parsed_args.force_all)
     checks_failed |= format_.run(*files, revisions=revisions, exit_on_failure=exit_on_failure)
     checks_failed |= flake8_.run(*files, revisions=revisions, exit_on_failure=exit_on_failure)
-    # pylint is slow, so always run pylint in incremental mode
-    checks_failed |= pylint_.run(
-        *files,
-        revisions=[] if default_mode else revisions,
-        exit_on_failure=exit_on_failure,
-    )
+    checks_failed |= pylint_.run(*files, revisions=revisions, exit_on_failure=exit_on_failure)
 
     # run typing and coverage checks
     exit_on_failure = not parsed_args.force_all
--- a/applications_superstaq/check/all_.py
+++ b/applications_superstaq/check/all_.py
@@ -47,7 +47,7 @@ def run(*args: str, parser: argparse.ArgumentParser = check_utils.get_file_parse
         action="extend",
         help="Run checks in incremental mode.",
     )
-    parsed_args = parser.parse_args(args)
+    parsed_args = parser.parse_intermixed_args(args)
     files = parsed_args.files
     revisions = parsed_args.revisions
 
--- a/applications_superstaq/check/all_.py
+++ b/applications_superstaq/check/all_.py
@@ -3,6 +3,7 @@
 import argparse
 import sys
 import textwrap
+from typing import Optional
 
 from applications_superstaq.check import (
     build_docs,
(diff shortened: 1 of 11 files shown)


richrines1:
one probably unrelated thing i just noticed (maybe not for the PR/not really super important in general): if i run something like,
```
/check/pytest_.py file_that_does_not_exist.py
```
it will proceed as if i passed no arguments (i.e. run every test)

perlinm:
Hmm... I'm not sure how to fix this actually.  At the moment, `check/pytest_.py` identifies any arguments that are files (i.e. arguments that pass `os.path.isfile`), extracts those into a `files` list, then passes all other arguments to `pytest`.  I thought the ability to pass along arguments would be nice.

What's happening here is that `check/pytest_.py` doesn't see `file_that_does_not_exist.py`, so it assumes that this was an argument to pass on to `pytest`.

One possibility is to decide that we don't pass special arguments to `pytest`, and declare that if a user wants to do this, then they should just call `pytest` directly instead of `check/pytest_.py`.  I think this sort of fix would be as simple as replacing [`parser.parse_known_intermixed_args`](https://github.com/SupertechLabs/applications-superstaq/blob/fd170b1a4f1db0682795cabbec86682bb41f56c1/applications_superstaq/check/pytest_.py#L54) with `parser.parse_intermixed_args` (both here and in other check scripts).

richrines1:
i also like the ability to pass arguments, so i think it's probably preferable to leave as is vs. fixing the nonexistent file thing. the other option i can think of is to add a positional "files" argument to the argument parser,
```
parser = ...
parser.add_argument("files", nargs="*")
# (then add other parser args...)
```
that way `parse_known_intermixed_args` will interpret anything passed before any flags (i.e. things starting with "-") as a file, and anything after any flags as an argument to be passed to pytest, e.g.
```
parser = argparse.ArgumentParser()
parser.add_argument("files", nargs="*")
parser.add_argument("-i", "--incremental", action="store_true")

args = ["file1.py", "file2.py", "--flag-for-pyest", "arg-for-pyest-flag", "-i", "--another-flag-for-pyest"]
parsed_args, pytest_args = arparser.parse_known_intermixed_args(args))

print(parsed_args)  # prints Namespace(files=['file1.py', 'file2.py'], incremental=True)
print(pytest_args)  # prints  ['--flag-for-pyest', 'arg-for-pyest-flag', '--another-flag-for-pyest']
```
then we can just throw an error for nonexistent paths in `files`

perlinm:
I like this solution @richrines1!  This should now implemented with the newly-revised [`extract_file_args` decorator](https://github.com/SupertechLabs/applications-superstaq/blob/85530281f0965a3f74143ea27264e8abf48f6e17/applications_superstaq/check/check_utils.py#L292).

However, I actually think the new implementation is a bit hackish and needlessly complicated.  For one, I no longer see any need for an explicit `file` argument to `check` scripts...  I think I'll remove both the `file` argument and the `extract_file_args` decorator.  Changes pending.

perlinm:
Alright I removed explicit `file` arguments from the `check` scripts, and there is now no need for the `extract_file_args` decorator.

applications_superstaq/check/check_utils.py
```diff
@@ -91,7 +92,7 @@ def should_include(file: str) -> bool:
 
 def get_changed_files(
     match_patterns: Iterable[str],
-    revisions: Iterable[str],
+    revisions: Optional[Iterable[str]],
```
richrines1, line 95:
```suggestion
    revisions: Optional[Iterable[str]] = None,
```

applications_superstaq/check/all_.py
```diff
@@ -17,12 +16,7 @@
 )
 
 
-@check_utils.extract_file_args
-def run(
-    *args: str,
-    files: Optional[Iterable[str]] = None,
-    parser: argparse.ArgumentParser = check_utils.get_file_parser()
-) -> int:
+def run(*args: str, parser: argparse.ArgumentParser = check_utils.get_file_parser()) -> int:
```
richrines1, line 19: should probably use something like
```suggestion
def run(*args: str, parser: Optional[argparse.ArgumentParser] = None) -> int:
    if parser is None:
        parser = check_utils.get_file_parser()
```
because ArgumentParser() is mutable

(ditto elsewhere)

richrines1 (commented):
nice, lg aside from a few nits!

applications_superstaq/check/check_utils.py
perlinm: Resolved by making the argument non-optional

applications_superstaq/check/all_.py
perlinm: I don't think I understand the issue.  Can you explain?

richrines1: >  Can you explain?

there are two issues i can think of:

first is pretty insignificant: all the `get_file_parser()` calls will happen as soon as the submodules are first imported instead of when a particular `run()` is called, which means we're doing the unnecessary work of instantiating all the parsers in check/*.py even if only one is actually being used (however (i think) these should be pretty lightweight so probably not a real problem in this case) 

second: because the resulting parser is mutable it's easy to end up with strange side effects. This is why pylist normally throws a "dangerous-default-value" error for mutable default arguments (apparently this only occurs for built-in types, which I suppose makes sense because pylint isn't a type parser). In this case, if you call `all_.run()` more than once it will raise and `ArgumentError: conflicting option string...` because the second call will try to add the same arguments to the default parser a second time. It's true that this shouldn't come up in normal use (because when run as a script `run()` won't be called more than once), but it's not portable and i can easily imagine it causing issues e.g. when we start adding checks for the checks

perlinm: Follow-up: should I rename `parser` to `_parser` to emphasize the fact that nobody calling `run` should actually pass in a parser?

perlinm:
Addition: I made `pylint_.py` incremental by default, with `-a/--all` options to truly check everything.  The real reason I did this is because `check/all_.py` was failing a complexity check, and this simplified some logic there.  But also I think it's just reasonable to do regardless.

applications_superstaq/check/all_.py
richrines1: > second issue: appears to not be the case.  Example: if I add a `return 0` after parsing args in `pytest_.py`, I can try `run(); run()`, which executes without errors.  I can also, say, `print(parser.description)` before and after defining it, and the prints behave as desired in both runs.

it doesn't seem to be an issue in pytest_.py, but it is in all_.py. For example if i add a line:
```
if __name__ == "__main__":
    print(run(*sys.argv[1:]))
    exit(run(*sys.argv[1:]))
```
the second `run()` raises `argparse.ArgumentError: argument -f: conflicting option string: -f`

perlinm: I see, I think it's an issue in `all_.py` because everywhere else I actually replace the parser with a new one inside the `enable_incremental` wrapper.  In any case, moot point now I think.

richrines1: > why decorator

i do really appreciate the decorator aesthetics, but i guess i think it ultimately ends up making both the *_ scripts and check_utils somewhat opaque/harder to follow: it kind of takes away the from "magically enable" functionality if you also have to add an extra unused argument in the function signature, and it's not really obvious what it's doing if you just read e.g. pytest_.py (i definitely wouldn't expect it to be using `inspect` to extract and update the default value of a function argument, which afaik isn't a typical practice in python)

so i guess imo making it a function would make it much simpler more obvious to future developers/maintainers what it's actually doing. But definitely ok with being overruled here; will approve (pending a few nits) if you think it should stay as is

richrines1:
> Addition: I made `pylint_.py` incremental by default, with `-a/--all` options to truly check everything if truly desired

seems reasonable to me

perlinm:
I am confused (failing incremental `pylint_.py` check on the server): 
image

This can be fixed by running `check/pylint_.py --all` in the github workflow, but that shouldn't be necessary...

richrines1:
we should still figure out what's causing the error, but imo the github action should use `--all`. pylint_ is for from the slowest check so it doesn't cost us anything in terms of time, and it would prevent unmodified files from being "grandfathered" in if e.g. we update pylint/its configuration to be more strict

perlinm:
Okay so @richrines1 I overhauled argument parsing altogether.  The format is now:
```python
parser = check_utils.get_file_parser()
...
parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)  # or something like this
files = check_utils.get_file_args(parsed_args, ...)  # takes care of incremental checks
...
```
For reasons unknown, the `pylint_.py` check now passes, but now the `pytest_.py` check gives really weird (seemingly git-related) errors.  Will continue to debug...

applications_superstaq/check/pytest_.py
```diff
@@ -51,30 +45,51 @@ def run(
...
+    if parsed_args.integration and integration_setup:
+        integration_setup()
+
+    return subprocess.call(["pytest", *files, *args_to_pass], cwd=check_utils.root_dir)
+
 
-    if parsed_args.integration:
-        if integration_setup:
-            integration_setup()
+def _get_file_search_options(
```
perlinm, line 66: Not sure whether there's a better way to do this, but I consider this much neater than what we had before.

applications_superstaq/check/all_.py
perlinm: [done](https://github.com/SupertechLabs/applications-superstaq/pull/92#issuecomment-1182534555)

richrines1 (commented):
nice! can we pass `--disable-socket` in coverage_.py as well?

(otherwise lgtm!)

perlinm:
done (also some nit cleanup/rearrangement of `coverage_.py`)

richrines1 (approved):
cool thanks! lgtm

PR #105, opened 2022-07-21
Add `typing.py` to `applications_superstaq`

### LINKED ISSUE #103: move TypedDicts to appplications-superstaq
as per dev meeting discussion: we will move the TypedDicts to appplications-superstaq

_Originally posted by @richrines1 in https://github.com/SupertechLabs/cirq-superstaq/pull/279#discussion_r921488703_

pgoiporia:
Fixes #103


perlinm:
Should **all** typing in `superstaq` go in here?  If so, we should throw [more types](https://github.com/SupertechLabs/SuperstaQ/blob/cq-transformers/superstaq/typing.py#L73-L98) into this PR (note: some of these I added in the `cq-transformers` branch of `superstaq` / [ss#1491](https://github.com/SupertechLabs/SuperstaQ/pull/1491))

pgoiporia:
> Should **all** typing in `superstaq` go in here? If so, we should throw [more types](https://github.com/SupertechLabs/SuperstaQ/blob/cq-transformers/superstaq/typing.py#L73-L98) into this PR (note: some of these I added in the `cq-transformers` branch of `superstaq` / [ss#1491](https://github.com/SupertechLabs/SuperstaQ/pull/1491))

I haven't moved all of them (for instance, the Qubit mapping TypedDict imports cirq, which we don't do in gss). But if the cq-transformers types don't require additional imports, I see no harm in moving them here/

vtomole:
>  I see no harm in moving them here

Agree. It does seem like all the types linked are specific to Cirq; hence can't be moved here.

vtomole (approved):
LGTM. Let's merge after @perlinm concurs that we haven't missed any non-Cirq/Qiskit Dependent types.

vtomole:
For now we don't use those Cirq types in `cirq-superstaq` but if we ever need to in the future then we can move those types to cirq-superstaq.

perlinm:
Alright since those types are all cirq-specific, we'll leave them out of `gss`.  In that case we can merge this.

Follow-up question: should we move the `cirq`-dependent types to `css` or just leave them in `ss`?

vtomole:
>  should we move the cirq-dependent types to css or just leave them in ss

Maybe we want to leave them in ss until the day the code in css needs them.

PR #110, opened 2022-07-26
make `check/plint_.py` incremental by default

perlinm:
#92 added an `-a/--all` flag to check everything with `check/pylint_.py`, but it looks like I didn't actually make `check/pylint_.py` incremental when called with no arguments.  Oops.

Also sneaking in colorizing the "Requirements files fixed." statement (making it green) after `./check/requirements.py --apply`, as otherwise you only get red text even though the problems have been fixed, which makes it look like there is still something wrong.


vtomole (approved):
LGTM

PR #112, opened 2022-07-26
Pylint cores issue

alexabuko:
This is a change which will make pylint_ run faster (by using multiple cores). Unfortunately, github only has two cores so it won't run faster on github, but this can at least let the checks run faster locally.

In order for the checks after changing pylint_ to work, setup.cfg had to be modified, so that change is reflected here as well.


applications_superstaq/check/pylint_.py
```diff
@@ -24,11 +26,18 @@ def run(
         """
     )
 
+    numCores = math.ceil(multiprocessing.cpu_count() / 2)
```
perlinm, line 29: Why `ceiling` instead of `multiprocessing.cpu_count() // 2` (which does a floor and doesn't require an import?

alexabuko: If someone has one core I want it to be set to 1 and not 0 (though we discovered that the autodetect functionality of pylint doesn't work and just uses one)

alexabuko: I don't think anyone would have one core in this day and age but it's more the "what if"

applications_superstaq/check/pylint_.py
```diff
@@ -24,11 +26,18 @@ def run(
...
+
     parser.add_argument("-a", "--all", action="store_true", help="Run pylint on the entire repo.")
+    parser.add_argument("-j", "--cores", type=int, nargs=1,
+                        default=[numCores], help="Number of cores to use for this test.")
 
     parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
     files = check_utils.extract_files(parsed_args, include, exclude, silent)
 
+    parsed_args.cores = parsed_args.cores[0]
+    args_to_pass.append("-j " + str(parsed_args.cores))
```
perlinm, line 39:
```suggestion
    args_to_pass.append(f"-j{parsed_args.cores}")
```

setup.cfg
```diff
@@ -58,15 +58,13 @@ enable=
...
     missing-kwoa,
-    mixed-indentation,
     mixed-line-endings,
     not-callable,
     no-value-for-parameter,
     nonexistent-operator,
     not-in-loop,
     pointless-statement,
     redefined-builtin,
-    relative-import,
```
perlinm, line 69: Why get rid of `mixed-indentation` and `relative-import`?

vtomole: I think this is related to https://github.com/SupertechLabs/SuperstaQ/pull/1431/files

alexabuko: I could not run the script because those two settings kept giving me errors. We can ignore that change if it's getting addressed elsewhere. I just forgot that I had changed that file. Sorry!

applications_superstaq/check/pylint_.py
perlinm: That's a good point.  How about `max(multiprocessing.cpu_count() // 2, 1)` then?

(I'm being super nitpicky; I just prefer to have a leaner dependency footprint if possible.)

alexabuko: Sure! I don't have a preference as long as my concern is taken care of!

setup.cfg
alexabuko: So should I just get rid of these changes for now then?

perlinm: I would be okay with sneaking the changes in [ss#1431](https://github.com/SupertechLabs/SuperstaQ/pull/1431) into this PR (just mention it in the first comment of this PR.

general_superstaq/check/pylint_.py
```diff
@@ -24,11 +25,24 @@ def run(
...
 
+    numCores = max(multiprocessing.cpu_count() // 2, 1)
+
     parser.add_argument("-a", "--all", action="store_true", help="Run pylint on the entire repo.")
+    parser.add_argument(
+        "-j",
+        "--cores",
+        type=int,
+        nargs=1,
+        default=[numCores],
```
perlinm, line 36:
```suggestion
        default=numCores,
```

general_superstaq/check/pylint_.py
```diff
@@ -24,11 +25,24 @@ def run(
...
+        help="Number of cores to use for this test.",
+    )
 
     parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
     files = check_utils.extract_files(parsed_args, include, exclude, silent)
 
+    parsed_args.cores = parsed_args.cores[0]
```
perlinm, line 43:
```suggestion
```

alexabuko: I did this because whenever I tried to input a value (in the terminal saying `./check/pylint_.py -j 4`) it would be in an array

perlinm: Is there not a way to make `parsed_args.cores` just an integer?

perlinm: I think it works if you remove the `nargs=1` line

alexabuko: It does! Thank you

perlinm:
@vtomole this looks good to me but I'm leaving it for your approval.

vtomole (approved):
LGTM

PR #113, opened 2022-07-27
Loosen package requirements

perlinm:
... to latest minor versions, e.g. `general_superstaq~=0.2.*` (for upstream `*superstaq*` packages only)


perlinm:
Good point.  Fixed.

richrines1 (approved):
lgtm!

PR #124, opened 2022-08-09
Add gpu simulation support to client

paaige:


general_superstaq/typing.py
```diff
@@ -54,5 +54,6 @@
         "data": Optional[Dict[str, Any]],
         "samples": Optional[Counter],
         "shots": Optional[int],
+        "result": Optional[Counter],
```
vtomole, line 57: Why are we adding this? Why can't we use `samples` here instead?

vtomole (changes requested):
Need to pass the checks.

vtomole (commented):
We need to add a `method` argument after https://github.com/SupertechLabs/SuperstaQ/pull/1500#pullrequestreview-1082958740

vtomole (approved):
LGTM

vtomole (commented):
Wait. Is options a string or a dict? https://github.com/SupertechLabs/cirq-superstaq/pull/312/files#r955130631

paaige:
It's a JSON serialized dictionary (a string)

PR #132, opened 2022-08-24
py3.7 compatibility for typeddict

### LINKED ISSUE #131: Make typing.py py37 compatible
I can't use css / qss with binder right now because they import gss.typing which has `TypingDict`, which is not available in python 3.7.

~I think the solution would be to add `from __future__ import annotations` in typing.py~

singular-value:
Closes #131


vtomole (approved):
LGTM. We might want to consider downgrading our Python versions to 3.7 given that it's Binder's default.

singular-value:
Tested locally to ensure it works with py3.7.

> LGTM. We might want to consider downgrading our Python versions to 3.7 given that it's Binder's default.

Good point. Let's discuss offline.

Issue #114, opened 2022-07-28
check that all `setup.cfg` files are in sync across repos


perlinm:
Right now our `setup.cfg` files in `gss/qss/css/ss` are almost identical (I believe except for one line).  It would be nice to have a check that makes sure these files are in sync.  @alexabuko we can talk about how this might work.

perlinm:
It looks like after #134, the only difference in `setup.cfg` across different repos will be a single line under `[coverage:run]`: `[gqc]ss/setup.cfg` have `omit = ./venv/*`, while `SuperstaQ/setup.cfg` has `omit = ./dev_tools/*`.

I think the `omit` line in SuperstaQ should be just removed, especially as it's non-functional.  Not sure about the `venv` one, although regardless of whether we keep it or toss it, we can make the choice identical across all repos.  That would make `setup.cfg` identical across all repos, which would greatly simplify this issue.  Having said that:
1. Whatever mechanism we use to ensure that `setup.cfg` is identical across all repos, we should build in a simple way of adding exceptions (otherwise it's a big :-1: for agility if and when we want to change a config in a single repo).  Maybe the checker could ignore lines with a special string (like, say, "# CUSTOM REPO CONFIG").
2. Do we want to keep the `venv` line in `[gqc]-ss`?  Not sure what everyone else's practices are here, but the disparity with SuperstaQ makes me suspect that this ignore is unnecessary?  @vtomole @richrines1

vtomole:
Will have to think more about what you are saying here but right off the bat I agree that the `omit` in SuperstaQ can be removed.

vtomole:
The `venv` one can be removed as well. Personally i don't run my `venv` in my root directory anymore. We can say on the dev guide that this is not allowed for us.

vtomole:
> `# REPO-SPECIFIC CONFIG`

I like this as an escape hatch.

PR #134, opened 2022-08-29
Add `isort` to `format_.py` and move `black` configs to `setup.cfg`

alexabuko:
This fixes [1321](https://github.com/SupertechLabs/SuperstaQ/issues/1321) in SuperstaQ. 

The `format_.py` check now runs both `black` and `isort`. (And both can automatically be fixed by adding the `--apply` tag)

Configurations for `black` are now also in `setup.cfg`.


general_superstaq/check/format_.py
```diff
@@ -30,12 +30,15 @@ def run(
...
+    isort_args_to_pass = []
     if not parsed_args.apply:
         args_to_pass = ["--diff", "--check"] + args_to_pass
+        isort_args_to_pass = ["--diff", "--check"]
 
     returncode = subprocess.call(["black", *files, *args_to_pass], cwd=check_utils.root_dir)
+    returncode += subprocess.call(["isort", *files, *isort_args_to_pass], cwd=check_utils.root_dir)
 
-    if returncode == 1:
+    if returncode > 1:
```
alexabuko, line 41: I don't know if this is the way I should do this. If either the black or isort don't pass their checks, they return 1, so then you want to tell the user to fix the issue. However, if they both don't succeed, the returncode would be 2. Is this okay?

perlinm: `black` has a return code of 1 iff it encounters formatting issues that it can fix, without any other errors.  Only in that case do we want to have the print statement inside this `if` (otherwise, rerunning with `--apply` won't actually fix anything).  For that reason, we want to keep this `returncode == 1`.

I would just separately keep track of a `returncode_black` and `returncode_isort`, and then also print the statement about rerunning with `--apply` when `isort` similarly only encounters formatting issues that it can fix.

perlinm: Now that I think of it, we probably don't even want to run `isort` if `black` has a return code other than `0` or `1`, so that's another thing to consider.

alexabuko: I can make those changes!

general_superstaq/check/format_.py
```diff
@@ -30,17 +30,25 @@ def run(
...
     if not parsed_args.apply:
         args_to_pass = ["--diff", "--check"] + args_to_pass
+        isort_args_to_pass = ["--diff", "--check"]
 
-    returncode = subprocess.call(["black", *files, *args_to_pass], cwd=check_utils.root_dir)
+    returncode_black = subprocess.call(["black", *files, *args_to_pass], cwd=check_utils.root_dir)
 
-    if returncode == 1:
+    if returncode_black > 1:
+        return returncode_black
```
perlinm, line 41:
```suggestion
    if returncode_black > 1:
        # this only occurs if black could not parse a file (for example due to a syntax error)
        return returncode_black
```

general_superstaq/check/format_.py
```diff
@@ -30,17 +30,25 @@ def run(
...
+
+    returncode_isort = subprocess.call(["isort", *files, *isort_args_to_pass], cwd=check_utils.root_dir)
+
+    if returncode_black == 1 or returncode_isort == 1:
```
perlinm, line 45: Does `isort` also only return `1` iff there are formatting issues?

general_superstaq/check/format_.py
```diff
@@ -30,17 +30,25 @@ def run(
     files = check_utils.extract_files(parsed_args, include, exclude, silent)
 
     args_to_pass = ["--color", "--line-length=100"] + args_to_pass
```
perlinm, line 32: nit: since we have two of these, let's call them `args_to_pass_black` and `args_to_pass_isort` for clarity

general_superstaq/check/format_.py
alexabuko: At least that is what I determined through my searching of documentation (though there was not a super clear answer)

alexabuko:
Does anyone know why isort needs the colorama package? Locally this check passes, and black doesn't have this issue, but I don't know why it fails online. Should I add colorama to my packages?

perlinm:
It might pass locally because you might have `colorama` installed locally.  You can check with `pip show colorama`.  In any case, you can just add `isort[colors]` to `dev-requirements.txt` (in this PR) to fix this.

general_superstaq/check/format_.py
perlinm: Let's test that.  Add a syntax error to some file in the repo, run `isort` on it, check the return code.  Then remove the syntax error, but add a "formatting error" (such as changing the order of imports), and again check the return code.

alexabuko:
I might not need to because I might not even be able to use the color option in the first place? I got this message: `WARNING: isort 5.10.1 does not provide the extra 'color'`

general_superstaq/check/format_.py
perlinm: `isort` doesn't fail due to syntax errors, so we should only ever get a return code of `1` if there is something to fix.  We can leave `== 1` as is.

alexabuko:
Sounds good to me!

alexabuko:
I can't find something which changes the import order for isort to be similar to what flake8 is currently set to (There seems to be an issue with order for items after a from but I can't figure out how to change that specific order). If anyone wants to help it would be greatly appreciated!

perlinm:
Looks like `flake8` and `isort` are in conflict?  This can probably be fixed by changing one of their settings.  For example, I think we want `isort` to be case-insensitive.  I'm also not sure about this `Application Relative` vs. `Application` business, but again I would err on deference to `flake8` by changing `isort` settings.

image

perlinm:
Looks like we can make `flake8` happy by appending the following under `[flake8]` in `setup.cfg`:
```
extend-ignore =
  I101,  # incorrect import name order (handled by isort)
  I202  # unexpected newlines in imports (handled by isort)
```

alexabuko:
I will do that! I just wasn't sure if there was a specific order we wanted to stick with (which is why I did not want to change anything in flake8)

setup.cfg
```diff
@@ -18,10 +24,6 @@ max-line-length = 100
 show_source = True
 statistics = True
 
-# specific configurations for flake8-import-order plugin:
-import-order-style = smarkets
-application-import-names = general_superstaq  # these are the libraries considered local
-
```
perlinm, line 24: Whoa does this mean that `setup.cfg` is now (or soon will be) identical in all repos?  That would make #114 much easier.

perlinm (commented):
LGTM.  Will wait for @vtomole or @richrines1 to give the final approval.  To that end, I'll mention here that the only non-formatting changes in this PR are to `check/format_.py`, `dev-requirements.txt`, and `setup.cfg`.

general_superstaq/check/format_.py
```diff
@@ -19,28 +19,38 @@ def run(
...
     parser.add_argument("--apply", action="store_true", help="Apply changes to files.")
 
-    parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
+    parsed_args, args_to_pass_black = parser.parse_known_intermixed_args(args)
     files = check_utils.extract_files(parsed_args, include, exclude, silent)
 
-    args_to_pass = ["--color", "--line-length=100"] + args_to_pass
-    if not parsed_args.apply:
-        args_to_pass = ["--diff", "--check"] + args_to_pass
+    args_to_pass_black = args_to_pass_black
```
vtomole, line 31: Why are we assigning a variable to itself here?

perlinm: Oops.  That's an artifact from changing the code a few times.  Fixed.

dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
 black[jupyter]~=22.3.0
+black-configparser~=0.2.0b2
```
vtomole, line 2: Why this addition?

perlinm: So that we can configure black in `setup.cfg` instead of passing arguments to it in `format_.py`

vtomole: Cool, let's keep the known issues with this package in mind (See the bottom of https://github.com/danie1k/python-black-configparser/blob/master/README.md)

```
Known issues

    Undefined behavior, when running with one of allowed CLI arguments which is also set in the config file.
    After black-configparser package is uninstalled, the black command does not work anymore and [black](https://pypi.org/project/black/) package must be reinstalled.
```

vtomole (approved):
LGTM

perlinm:
I'll leave it to @alexabuko to do the final merge :slightly_smiling_face:

richrines1:
> Looks like we can make `flake8` happy by appending the following under `[flake8]` in `setup.cfg`:
> 
> ```
> extend-ignore = I101, I202  # ignore import-related error codes, deferring to isort formatting
> ```
> 
> Let's just do that.
> 
> EDIT: haha nevermind. The actual fix should be to remove `flake8-import-order` from `dev-requirements.txt` (and to uninstall it locally with `pip uninstall flake8-import-order`).

@perlinm why did we decide against the extend-ignore fix? ideally the tests shouldn't fail just because a user happens to have `flake8-import-order` installed

perlinm:
Also `flake8-import-order` is replaced by `isort`

richrines1:
yeah i get that, but if i happen to have flake8-import-order installed locally (for unrelated reasons) it still causes /check/flake8_ to fail. but there doesn't seem to be any way around that short of covering all possible extension error codes in extend-ignores

perlinm:
Hmm, I'm not sure this makes sense to me.  As I see it, **not** having flake8-import-order installed is essentially just a dev-requirement for the `*superstaq` venvs.  Ideally, we could add a line like `!flake8-import-order` to ` dev-requirements.txt`, but to my (limited) knowledge pip doesn't support "negative dependencies".

PR #143, opened 2022-09-06
Add script to check consistency of `setup.cfg`

### LINKED ISSUE #114: check that all `setup.cfg` files are in sync across repos

perlinm:
fixes #114


perlinm:
Minor question: do we even need to add `check/configs.py` to other repos?
Pro: consistency with the fact that copy every other `check` script, and it would of course let us run `check/configs.py` manually.
Con: clutter in the `check` directory.  I don't **really** see myself ever running `check/configs.py`, and this is already run by `check/all_.py`.

I'm more sold by the cons myself.

richrines1:
how does "REPO-SPECIFIC CONFIG" work if i want to change an existing line in setup.cfg? e.g. if i want to either remove the `no_implicit_optional = true` flag or change it to `no_implicit_optional = false`, it seems like checking for `REPO-SPECIFIC CONFIG` won't work because it won't match in `gss`'s setup.cfg

perlinm:
Good point.  Let me think on this.  Ideas from others also welcome.

perlinm:
@richrines1 solved.  If you don't want, say, `no_implicit_optional = true`, you can comment it out with `# no_implicit_optional = true  # REPO-SPECIFIC CONFIG`.

general_superstaq/check/configs.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if not deltas:
+        if not silent:
+            print(check_utils.success("Config files are consistent!"))
+        return 0
+
+    # print diffs
+    print(check_utils.failure(f"Found {len(deltas)} differences between config files:"))
+    print(f"{file_orig} (original)")
+    print(f"{file_copy} (copy)")
```
richrines1, line 54: nit
```suggestion
    print(check_utils.styled(f"< {file_orig} (original)", check_utils.Style.RED))
    print(check_utils.styled(f"> {file_copy} (copy)", check_utils.Style.GREEN))
```
to match the text below (similar to the output of `diff`)

general_superstaq/check/configs.py
```diff
@@ -0,0 +1,89 @@
...
+    for tag, orig_start, orig_end, copy_start, copy_end in deltas:
+        _announce_diff(tag, orig_start, orig_end, copy_start, copy_end)
```
richrines1, line 56: does _anounce_diff present any info that isn't obvious from the text below?

general_superstaq/check/configs.py
```diff
@@ -0,0 +1,89 @@
...
+) -> None:
+    assert tag in ["replace", "delete", "insert"]
+    line_text_orig = _line_text(orig_start, orig_end)
+    line_text_copy = _line_text(copy_start, copy_end)
+    if tag == "replace":
+        text = f"replaced {line_text_orig} of original by {line_text_copy} of copy"
+    elif tag == "delete":
+        text = f"deleted {line_text_orig} from original"
+    elif tag == "insert":
+        text = f"inserted {line_text_copy} of copy"
```
richrines1, line 78: fine to leave as is if you disagree, but this grammar feels a little off to me - it seems like it should be something more like "line x of copy not present in original" (or ideally something less verbose) (and similarly for the above text)

richrines1 (commented):
few nits, otherwise lg!

general_superstaq/check/configs.py
perlinm: `_announce_diff` is responsible for the cyan lines here: 
image
e.g. "inserted lines 12--14 of copy".  Since `setup.cfg` files need to be fixed manually, it's helpful to have the check script tell us where to look.

general_superstaq/check/configs.py
perlinm: New grammar:
image

general_superstaq/check/configs.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _announce_diff(
+    tag: str, orig_start: int, orig_end: int, copy_start: int, copy_end: int
+) -> None:
+    assert tag in ["replace", "delete", "insert"]
+    line_text_orig = _line_text(orig_start, orig_end)
+    line_text_copy = _line_text(copy_start, copy_end)
+    if tag == "replace":
+        text = f"{line_text_orig} of original have been replaced by {line_text_copy} of copy"
```
richrines1, line 75:
```suggestion
        text = f"{line_text_orig} of original replaced by {line_text_copy} of copy"
```
(to avoid tense mismatch for single-line changes, e.g. "line 3 of original have")

richrines1 (approved):
one more nit then looks good!

richrines1 (approved):

PR #144, opened 2022-09-06
don't check cirq and qiskit versions

### LINKED ISSUE #130: remove cirq warning from check.requirements
in `css` we usually try to maintain support for a previous version or two of cirq, so the warning is a little misleading

perlinm:
reverts #121
addresses (but does not fully fix) #130


richrines1 (approved):
lgtm

PR #146, opened 2022-09-07
"backend" -> "target" in json content

richrines1:
use `target` whenever referring to a superstaq-formatted target string (e.g. "ibmq_lima_qpu")

review downstream PRs before merging


dowusu-antwi (changes requested):
nits in `finance.py` lines 64-68: we can change L64 from
```python
"""Submits the given QUBO to the target backend. The result of the optimization
```
to
```python
"""Submits the given QUBO to the target. The result of the optimization
```
and L68 from
```python
target: A string indicating which backend to use.
```
to 
```python
target: A string indicating which target to use.
```

dowusu-antwi (approved):
lgtm after nits

richrines1:
> nits in `finance.py`

good catches. done.

PR #150, opened 2022-09-09
GitHub checks cache

alexabuko:
These changes are related to issue [#365 ](https://github.com/SupertechLabs/SuperstaQ) in SuperstaQ, which need to be made before the cache can be implemented.

How it works:
The `date` job is used in all of the tests, hence any subsequent job uses the `needs: date` line. The date job gets today's date to use in the cache keys. Then, for the tests, the steps `Get folder locations` and `Cache pip installs` are added after the python install step to generate/load the cache.

Changes that have been made:
The `make-cacke.yml` file generates the github cache daily. The check file (`ci.yml`) has also been updated to use this generated cache. The original check file has still been kept and renamed to `ci-no-cache.yml` which can me manually run.

What needs to be done:
- In theory, the daily cache script should work, but more testing should be conducted before this is implemented. It seems to have worked for all tests consistently except `format` (though this should work). The `pytest windows` test does not work with the cache, so this test has not been sped up.
- A functionality to reset a day's cache should also be nice. This link might be helpful: https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key

If anything needs clarifying, feel free to send me a message!


stephanielee9:
@PalashGoiporia to look over later

PR #151, opened 2022-09-20
Update the return type of `convert_qubo_to_model`

vtomole:
Fixes https://github.com/SupertechLabs/general-superstaq/issues/29


vtomole:
@hallben4 Please review.

hallben4 (approved):
lgtm

vtomole:
Thanks!!

general_superstaq/qubo.py
```diff
@@ -16,7 +16,7 @@ def read_json_qubo_result(json_dict: dict) -> np.recarray:
     return gss.converters.deserialize(json_dict["solution"])
 
 
-def convert_qubo_to_model(qubo: qv.QUBO) -> List[Dict[str, Any]]:
+def convert_qubo_to_model(qubo: qv.QUBO) -> List[Dict[str, Iterable[str]]]:
```
richrines1, line 19: is this annotation correct? i think the "value" entry is just a number, not an iterable

richrines1: also out of curiosity why isn't our model just `dict(qubo)`

vtomole: It's not correct.  That's why we didn't do a deployment last week. I meant to follow up on this but didn't get to.

richrines1: ah cool sorry still out of the loop. also realizing `dict(qubo)` doesn't work because it isn't json serializable

PR #153, opened 2022-09-28
Update the return type of `convert_qubo_to_model` part 2

vtomole:
Fixes https://github.com/SupertechLabs/general-superstaq/issues/29


general_superstaq/qubo.py
```diff
@@ -1,10 +1,13 @@
-from typing import Dict, Iterable, List
+from typing import TYPE_CHECKING, Dict, List
 
 import numpy as np
 import qubovert as qv
 
 import general_superstaq as gss
 
+if TYPE_CHECKING:
```
richrines1, line 8: at least locally everything seems to pass without this `if`, if you also remove the quotes from "QuboDict" below

vtomole: Wasn't explicitly importing the typing QubotDict. Fixed.

richrines1:
i think the "qubo" entries in the other gss.typing TypedDicts should also be changed to `List[QuboDict]`

perlinm (approved):
lgtm

richrines1:
> lgtm

ditto

PR #159, opened 2022-09-29
`QscoutDict`

vtomole:
Needed for https://github.com/SupertechLabs/qiskit-superstaq/pull/234

Will have @hallben4 review and @richrines1  review once approved.


general_superstaq/typing.py
```diff
@@ -68,3 +68,7 @@
         "shots": Optional[int],
     },
 )
+
+SerializedQscoutDict = TypedDict(
+    "SerializedQscoutDict", {"qiskit_circuits": str, "jaqal_programs": List[str]}
```
richrines1, line 73: this type is only relevant to `qss` so imo should be moved there -- the same endpoint returns `"cirq_circuits"` or `"qasm_strs"` when access with different clients/input format (and atm `gss` doesn't contain any references to qiskit/cirq)

vtomole: Cool, will do that once generics are disabled for `gss`.

stephanielee9:
Absorbed into https://github.com/SupertechLabs/general-superstaq/pull/167

PR #162, opened 2022-09-30
don't build docs if there is no `docs/Makefile`

### LINKED ISSUE #161: `./check/all_.py` fails in repos without a `docs/Makefile`
There is no reason for this to fail: 
image

Specifically, it's failing at the `build_docs` stage.

perlinm:
fixes #161


vtomole (approved):
LGTM

PR #164, opened 2022-10-01
Add `setup.cfg` to `gss` package, and manually retrieve package root repo

### LINKED ISSUE #163: check/configs.py only works when gss is installed locally
if gss was installed via pip/anywhere other than a git repository, check/configs.py (and therefore check/all_.py in any repo) will fail with:
```
subprocess.CalledProcessError: Command '['git', 'rev-parse', '--show-toplevel']' returned non-zero exit status 128.
```
due to [this line](https://github.com/SupertechLabs/general-superstaq/blob/f1f147860293beb72c45240d53e28c707d0a12ec/general_superstaq/check/configs.py#L31) (which doesn't work because it's just a folder in site-packages, not a git repo). Fixing this raises another issue, because `setup.cfg` currently isn't copied into `site-packages/general_superstaq` (which i think can be fixed in setup.py)

perlinm:
fixes #163, I think -- I'm not sure how to test `setup.py`.

Note: I did not figure out how to retrieve the package root directory automagically, so I hard-coded it as two directories up from `general_superstaq/check/configs.py`.  Better ideas welcome.


setup.py
```diff
@@ -43,5 +43,5 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     packages=general_superstaq_packages,
-    package_data={"general_superstaq": ["py.typed"]},
+    package_data={"general_superstaq": ["py.typed"], "": ["setup.cfg"]},
```
richrines1, line 46: where will this go? when i install gss from pypi it only includes the `general_superstaq` subdirectory in site-packages, which would mean this would have to be at `site_packages/setup.cfg` (which definitely seems wrong)

maybe setup.cfg could be moved to `general_superstaq/check` and then just symlinked in the root directory?

perlinm: I like moving to the check script and symlinking.  Possibly not Windows friendly, but I'm not sure whether that's a concern since I think our Windows developers use WSL.

perlinm: done

richrines1 (approved):
lgtm!

PR #166, opened 2022-10-04
ignore flake8 import order

richrines1:
prevents flake8 from failing if `flake8-import-order` happens to be installed (due to #134 we enforce a different import order via `isort`)


pgoiporia (approved):
LGTM

Issue #158, opened 2022-09-29
Create extra flag [examples] for qss and css


vtomole:
This will install all the requirements needed to run the notebooks in `examples`

For example

`pip install cirq-superstaq[examples]`

Should install:

```
notebook
pulser
qiskit # only for css
```

@dowusu-antwi Please confirm that this will solve https://github.com/SupertechLabs/cirq-superstaq/issues/136

dowusu-antwi:
> @dowusu-antwi Please confirm that this will solve [SupertechLabs/cirq-superstaq#136](https://github.com/SupertechLabs/cirq-superstaq/issues/136)

Yep, this subsumes https://github.com/SupertechLabs/cirq-superstaq/issues/136.

vtomole:
@perlinm The reason we are doing this flag approach to avoid having an `examples/requirements.txt` in every repo.

perlinm:
I guess I was imagining a situation with `examples` having project sub-directories (such as `examples/qiskit-runtime` or `exmples/qiskit-runtime`), and each project defining its own requirements.  Sounds like that's not what you have in mind though.

knsmith:
https://github.com/SupertechLabs/qiskit-superstaq/pull/245 and https://github.com/SupertechLabs/cirq-superstaq/pull/358 will solve this issue.

knsmith:
https://github.com/SupertechLabs/qiskit-superstaq/pull/245 and https://github.com/SupertechLabs/cirq-superstaq/pull/358 have been merged.

PR #170, opened 2022-10-06
Create example flag qss css

knsmith:
Solves issue #158 .

Added example-requirements.txt and edited setup.py


example-requirements.txt
```diff
@@ -0,0 +1,5 @@
+
```
vtomole, line 1: Nit: remove newline.

example-requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+notebook
+pulser
+qiskit>=0.35.0, <0.37.0
+
```
vtomole, line 5: Nit: remove newline.

vtomole (commented):
Looks good modulo nits. After merge the next step is to add READMEs to the  `examples` directories in qss and css telling users to install the example requirements.

example-requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+notebook
+pulser
```
vtomole, line 3: Pin the requirement
```suggestion
pulser~=0.7.0
```

example-requirements.txt
```diff
@@ -0,0 +1,5 @@
+
+notebook
```
vtomole, line 2: Pin requirement
```suggestion
notebook~=6.4.12
```

richrines1:
nice! but should this be moved to `qss` & `css`? afaik general_superstaq doesn't have examples

perlinm:
Dropping in with two questions:
1. The empty lines in `example-requirements.txt` should have raised errors from `check/requirements.py`.   If this PR was passing checks with the empty lines, can we look into why that check *wasn't* failing?
2. What do we do in the hypothetical case that some examples have conflicting requirements?  At some point in the past there was an example that used `qiskit-runtime`, which was incompatible with some of the non-example SuperstaQ requirements.  I suppose an answer could be that we don't throw that example's requirements in `example-requirements.txt`, but I wanted to pitch this to others for their thoughts.

knsmith:
Circling back to the discussion here. I added Victory's changes (removing empty lines) and checked out @perlinm 's comments. Are there any action items I should take on?

vtomole:
> moved to `qss` & `css`

See https://github.com/SupertechLabs/general-superstaq/issues/158#issuecomment-1271959037, but the fact that we are installing qiskit twice in `qss` makes me want to revisit this decision.

vtomole:
>  I suppose an answer could be that we don't throw that example's requirements in example-requirements.txt

This sounds good to me.

richrines1:
> See [#158 (comment)](https://github.com/SupertechLabs/general-superstaq/issues/158#issuecomment-1271959037), but the fact that we are installing qiskit twice in `qss` makes me want to revisit this decision.

i agree. imo the point of the flag is to be able to use `pip install cirq(qiskit)-superstaq[examples]` to install the requirements for cirq(qiskit)-superstaq examples. i'm not sure it really makes sense to put this in gss because (1) the two don't have the same example requirements, (2) the requirements are unrelated to anything in gss itself, (3) we'll still need to update both qss and css to install `general-superstaq[examples]` when the `[examples]` flag is passed, so it doesn't really save anything

vtomole:
@knsmith  See @richrines1's latest comment.

richrines1:
>     2. What do we do in the hypothetical case that some examples have conflicting requirements?  At some point in the past there was an example that used `qiskit-runtime`, which was incompatible with some of the non-example SuperstaQ requirements.  I suppose an answer could be that we don't throw that example's requirements in `example-requirements.txt`, but I wanted to pitch this to others for their thoughts.

mostly agree as well

another answer is that we maybe shouldn't have "examples" with conflicting requirements, because it can't be installed alongside `*superstaq` we can't really call it an example of `*superstaq` in the first place. qiskit-runtime is maybe an exception, but i'd kinda consider this a bug (which should be fixed once we can merge https://github.com/SupertechLabs/qiskit-superstaq/pull/207)

knsmith:
Closing this PR because https://github.com/SupertechLabs/qiskit-superstaq/pull/245 and https://github.com/SupertechLabs/cirq-superstaq/pull/358 will solve https://github.com/SupertechLabs/general-superstaq/issues/158

PR #172, opened 2022-10-07
handle aqt configs as dictionaries

richrines1:
* allow dictionaries of configuration data to be passed to `aqt_upload_configs()`, in addition to file paths
* return config dictionaries from `aqt_download_configs()` when no file paths are specified

some functionality requires pyyaml to be installed. Because (1) it's unnecessary outside of these two endpoints and (2) users of these endpoints will almost certainly have pyyaml installed already, i left it out of requirements.txt and instead raise an error when it's required but not available (however for the purposes of testing it is included in dev-requirements.txt)


hallben4 (approved):
lgtm

PR #179, opened 2022-10-20
Upgrade pytest version

### LINKED ISSUE #178: Upgrade pytest version
The current version of pytest (6.1.2) results in an error TypeError: required field "lineno" missing from alias, related to [this known issue](https://github.com/pytest-dev/pytest/discussions/9195). This issue was fixed in version >= 6.2.5.

paaige:
Closes #178


perlinm (approved):

dev-requirements.txt
```diff
@@ -5,7 +5,7 @@ isort[colors]~=5.10.1
 mypy>=0.961
 nbmake~=1.3.0
 pylint>=2.13.0
-pytest~=6.1.2
+pytest>=6.2.5
 pytest-cov~=2.11.1
 pytest-randomly~=3.10.1
 pytest-socket~=0.4.1
```
richrines1, line 11: should we change these to `>=` as well?

paaige: I guess this only matters if going below these versions break things?

richrines1: `~=` should still prevent going below these versions, but does prevent us from using more recent ones (basically `~=2.11.1` is synonymous with `>=2.11.1, <2.12.0`). So we'd be allowing any future update of pytest, but pinning (already out-of date) versions of its plugins

perlinm: I vote in favor of `>=` if it doesn't break anything.

PR #182, opened 2022-10-21
correctly parse aqt configs from qtrl managers

richrines1:
fixes: #181 

requires users to pass the config managers themselves rather than casting them to dicts, e.g.:
```
client.aqt_upload_configs("pulses.yaml", "variables.yaml")  # still works
client.aqt_upload_configs(dict(schedule_config.pulses), dict(schedule_config.variables))  # still fails
client.aqt_upload_configs(schedule_config.pulses, schedule_config.variables)  # works after this pr
```


general_superstaq/user_config.py
```diff
@@ -33,45 +33,46 @@ def ibmq_set_token(self, token: str) -> Dict[str, str]:
...
         return self._client.ibmq_set_token({"ibmq_token": token})
 
-    def aqt_upload_configs(
-        self,
-        pulses: Union[str, Dict[str, Any]],
-        variables: Union[str, Dict[str, Any]],
-    ) -> Dict[str, str]:
-        """Uploads configs for AQT. Arguments can either be dictionaries or paths to valid
-        .yaml files. If neither is provided, the existing configuration is preserved.
+    def aqt_upload_configs(self, pulses: Any, variables: Any) -> Dict[str, str]:
```
singular-value, line 36: should we have `Union[qtrl.managers.PulseManager, str]` and `Union[qtrl.managers.VariableManager, str]` here instead of the Any's?

general_superstaq/user_config.py
```diff
@@ -33,45 +33,46 @@ def ibmq_set_token(self, token: str) -> Dict[str, str]:
...
+                try:
+                    import yaml
+
+                    return yaml.safe_dump(config)
+                except ImportError:
+                    raise ModuleNotFoundError(
+                        "The PyYAML package is required to upload AQT configurations from dicts. "
+                        "You can install it using 'pip install pyyaml'."
+                    )
+                except yaml.YAMLError:
```
singular-value, line 66: Is the intent to direct this error to be caught be L69 ValueError (which hopefully is a more helpful error message)? Seems fine but just confirming.

singular-value (approved):
Lgtm

Couple of minor questions, but fine with pushing too.

general_superstaq/user_config.py
richrines1: technically that's correct, but it would mean adding qtrl as a dependency for `gss`

general_superstaq/user_config.py
richrines1: yep - otherwise the error is basically a wall of text

general_superstaq/user_config.py
singular-value: 👍

general_superstaq/user_config.py
singular-value: 👍

PR #187, opened 2022-10-31
Update issue forms

stephanielee9:


.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yaml
```diff
@@ -1,5 +1,6 @@
 name: Feature request
-about: Suggest an idea for this project 🦅🐆
+description: Suggest an idea for this project 🦅🐆
+labels: ["enhancement"]
```
pgoiporia, line 3: Is there a reason we're using "enhancement" as a label rather than "feature request".

pgoiporia (approved):
LGTM! Just a quick nit. Approving and leaving it to you whether or not to change it.

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yaml
stephanielee9: I'll update - the reason I had enhancement is that it's one of the labels that had been used before

Issue #165, opened 2022-10-04
Run coverage checks for python 3.7, 3.9 and 3.10 on Github Actions


vtomole:
To catch Python dependency incompatibilities.

knsmith:
We will also want to add Python 3.11. https://github.com/actions/python-versions/releases

knsmith:
qiskit-superstaq will only run coverage checks for Python 3.7-3.10 until we start using qiskit>=0.37.0

knsmith:
Completed with:
https://github.com/SupertechLabs/general-superstaq/pull/189
https://github.com/SupertechLabs/qiskit-superstaq/pull/266
https://github.com/SupertechLabs/cirq-superstaq/pull/379
https://github.com/SupertechLabs/SuperstaQ/pull/1809

PR #189, opened 2022-11-07
add coverage checks for Python 3.7-3.11

knsmith:
Addresses #165 for general-superstaq


richrines1:
maybe a silly question, but is testing with python 3.7 and 3.10 sufficient to assume everything works w/ 3.8 and 3.9 as well?

knsmith:
@richrines1 I'm not sure...I guess we were including all tests to ensure no surprises? Thoughts, @vtomole? We'll include Python 3.11 as well since it's been released.

vtomole (changes requested):
Turns out we don't need to copy paste. We can have all the versions be in one check. See https://github.com/quantumlib/Cirq/blob/6fba8becd6d0fad05cfa7558ef5b1377b7a1d5b9/.github/workflows/ci.yml#L143

knsmith:
This will look so much better! Yay!

.github/workflows/ci.yml
```diff
@@ -79,14 +79,17 @@ jobs:
           check/mypy_.py
 
   coverage:
-    name: Pytest and Coverage check
+    name: Pytest and Coverage check for Python 3.7 - 3.11
```
vtomole, line 82: Lets revert this. The actions will show it anyway.
```suggestion
    name: Pytest and Coverage check
```

.github/workflows/ci.yml
```diff
@@ -79,14 +79,17 @@ jobs:
...
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.8
+      - name: Set up Python 3.7, 3.8, 3.9, 3.10, 3.11
```
vtomole, line 89:
```suggestion
      - name: Set up Python
```

vtomole (changes requested):

knsmith:
@vtomole , did I fix all of your requested changes?

vtomole (approved):
Yep thank you so much!

vtomole (changes requested):
Oh actually one more thing

vtomole:
Change https://github.com/SupertechLabs/general-superstaq/blob/8f7c58a88f6c360f49002769fab921471d8bfeb2/setup.py#L38 from 3.6 to 3.7.

vtomole (approved):
LGTM

PR #198, opened 2022-12-07
Implement `encode_files`

vtomole:


vtomole:
@PalashGoiporia Since we only use this function twice. I'm fine with using it in the notebooks.

PR #201, opened 2022-12-14
make check/setup.py return a warning

### LINKED ISSUE #185: make `./check/configs.py` raise a warning instead of an error
(Issue brought up by @richrines1 at a dev meeting)
Raising an error is a pain for when we make non-trivial changes to `setup.cfg` in `gss`.  A warning is sufficient for keeping us on our toes.

Question to @vtomole @richrines1: is there a "proper" way to raise a warning?  Or should we just make `./check/configs.py` yell at us but return a succeeding exit code of 0?

richrines1:
fixes #185


general_superstaq/check/check_utils.py
```diff
@@ -46,7 +45,7 @@ def styled(text: str, style_code: str) -> str:
 
 
 def warning(text: str) -> str:
-    return styled(text, Style.RED)
+    return styled(text, Style.BOLD + Style.YELLOW)
```
richrines1, line 48: (imo the red text was too easy to mistake for an error)

perlinm (approved):
lgtm

PR #203, opened 2022-12-21
use `post_request` in `ibmq_set_token`

richrines1:
(nit change to remove redundancy)


vtomole:
Still haven't figured out why gss doesn't autoassign but you can manually assign @SupertechLabs/code-reviewers for now.

vtomole (approved):

PR #207, opened 2023-01-09
Checking docstrings with flake8-docstrings

pgoiporia:
https://github.com/SupertechLabs/SuperstaQ/issues/696

flake8-docstrings seems to be a bit of overkill, so I'm overriding some of the errors. I might be accidentally skipping some, so I'm compiling a list of the skipped errors and what they mean

D100 - Every file is required to have a docstring at the top.
D107 - Every init is required to have a docstring.
D401 - First line should be in imperative mood; try rephrasing. (useful but too many exceptions.)

"# noqa" used to ignore flake8 on specific lines.


pgoiporia:
flake8-docstrings is an extension, not a separate package. So running `flake8` will automatically run `flake8-docstrings` if it is installed. This prevents from an incremental implementation. So instead, I propose not updating the requirements to include flake8-docstrings until I have merged in PRs to change docstrings in all the repos. Only then shall I update the requirements. If someone has a better workaround, please let me know.

vtomole:
It's better if we ignore the errors and make an issue to fix them. This makes sure that future malformed docstrings aren't let through. See https://github.com/SupertechLabs/SuperstaQ/pull/1885 for example.

pgoiporia:
> It's better if we ignore the errors and make an issue to fix them. This makes sure that future malformed docstrings aren't let through. See [SupertechLabs/SuperstaQ#1885](https://github.com/SupertechLabs/SuperstaQ/pull/1885) for example.

Thinking about this more, there are about 1100 flake8-docstrings failures in SuperstaQ right now. Putting in that many ignores seems a bit counterproductive. I think it's ok to not do it incrementally, as docstrings are added to the codebase at a pretty low rate. Alternatively, we could switch off flake8 as a whole and reintroduce it incrementally. Thoughts?

vtomole:
> 1100 flake8-docstrings

Whoa! Do we get that many errors with [docstyle](https://github.com/SupertechLabs/SuperstaQ/issues/696#issuecomment-1375894837) as well?

pgoiporia:
swtiching to pylint docstrings instead.

PR #210, opened 2023-01-11
Removing `# pylint: disable` flags.

pgoiporia:
Using [docstyle](http://pylint.pycqa.org/en/latest/technical_reference/extensions.html#pylint-extensions-docstyle) and [docparams](http://pylint.pycqa.org/en/latest/technical_reference/extensions.html#pylint-extensions-docparams) plugins.


general_superstaq/check/check_utils.py
```diff
@@ -40,19 +40,19 @@ class Style(str, enum.Enum):
...
     return styled(text, Style.BOLD + Style.YELLOW)
 
 
-def failure(text: str) -> str:
+def failure(text: str) -> str:  # pylint: disable=missing-function-docstring
     return styled(text, Style.BOLD + Style.RED)
 
 
-def success(text: str) -> str:
+def success(text: str) -> str:  # pylint: disable=missing-function-docstring
```
SalahedeenIssa, line 55: What exactly do these pylint docstrings signify?

general_superstaq/finance.py
```diff
@@ -56,7 +60,7 @@ def read_json_maxsharpe(json_dict: MaxSharpeJson) -> MaxSharpeOutput:
     return MaxSharpeOutput(best_portfolio, best_ret, best_std_dev, best_sharpe_ratio, qubo)
 
 
-class Finance:
+class Finance:  # pylint: disable=missing-class-docstring
```
SalahedeenIssa, line 63: Just for my own understanding, why are we disabling these?

pgoiporia: I'm disabling them because I'm not sure how to document the Finance class. The idea is that whoever wrote the class/understands it will take out the disable comment and add a docstring.

general_superstaq/check/check_utils.py
pgoiporia: See below

general_superstaq/_version_test.py
```diff
@@ -4,6 +4,7 @@
 
 
 def test_version() -> None:
+    """Test for general superstaq version."""
```
vtomole, line 7: Let's disable all needed docstrings for this PR

pgoiporia: how come?

general_superstaq/finance_test.py
```diff
@@ -57,6 +59,7 @@ def test_read_json_maxsharpe() -> None:
     },
 )
 def test_service_submit_qubo(mock_submit_qubo: mock.MagicMock) -> None:
+    """Test for submitting QUBO."""
```
vtomole, line 62: Is there a way for us to disable docstrings for test functions only? They tend to be redudant to what the test calls? Let's find out how Cirq disables test files.

general_superstaq/_version_test.py
vtomole: Just to make this PR small/do one thing: What we want to accomplish here is adding the doc check. Actually writing the documentation should be for future PRs. You can move the changes you have here to a different branch and create a PR for them after this one is merged.

pgoiporia: makes sense.

general_superstaq/finance_test.py
pgoiporia: I think we can just do a string matching on files. I'll check what they do though

pgoiporia: cirq seems to be have it's own infrastructure for sorting through modules and selecting files. See https://github.com/quantumlib/Cirq/blob/master/dev_tools/modules.py

pgoiporia: I'm going to use pylint ignores for now and then dig a bit deeper. Also, I believe if you put a pylint ignore at the top of the file, it should ignore for all test functions.

stephanielee9:
Needs to be reworked

PR #212, opened 2023-01-18
updated file extraction routine for tests

richrines1:
a few changes in anticipation of the repo merge:
* enables `check/<test_script>.py <subdir>` syntax for running `<test_script>.py` in a single subdirectory
* adds an `-x`/`--exclude` flag to all check scripts
* refactors a few file-extration-related functions in check_utils to better separate responsibilities
* fixes a bug where `-i`/`--incremental` mode wouldn't actually be incremental if no relevant file had changed
* fixes `pytest_.py -i` so it runs tests for any changed file (previously it would only include tests if the `x_test.py` file itself had changed)


general_superstaq/check/coverage_.py
```diff
@@ -32,7 +32,7 @@ def run(
     files = check_utils.extract_files(parsed_args, include, exclude, silent)
 
     silent = silent or not (parsed_args.files or parsed_args.revisions)
-    test_files = check_utils.get_test_files(*files, exclude=exclude, silent=silent)
+    test_files = check_utils.get_test_files(*files, silent=silent)
```
singular-value, line 35: @SalahedeenIssa and I are pair-reviewing this.

We're trying to understand why the `exclude` was deleted. Would it cause us to run integration tests in coverage? (since there is a default value for `exclude` in L15)

richrines1: ah i probably shouldn't have. just put it back.

it shouldn't matter for integration tests because all the include/exclude filtering is already done in `extract_files`. `get_test_files` just takes its input files and returns a corresponding list of test files (if they exist), so as long as there are no integration tests in `files` they won't be included in `test_files`*

the case where it does make a difference is if we want to exclude a specific \*non-integration\* test without excluding the corresponding source file (e.g. maybe we want to check if we still get complete coverage of `qubo.py` if we exclude `qubo_test.py`). Which is maybe unlikely but i think it's fair enough to expect this to work

*one exception is if for some reason we named a file something like `x_integration.py`, in which case it might return `x_integration_test.py` (which imo wouldn't really be "wrong")

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
...
+        dest="force_formats",
+        help="'Soft force' ~ continue past (i.e. do not exit after) failing format checks.",
+    )
+    parser.add_argument(
+        "-F",
+        "--force",
+        action="store_true",
+        dest="force_all",
+        help="'Hard force' ~ continue past (i.e. do not exit after) all failing checks.",
+    )
```
perlinm, line 228: We move these options here?  They are only used for `./check/all_.py`, and don't make sense for, say, `check/flake8_.py`.  Moving these options here would make them show up in, say, `check/flake8_.py -h`.

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
...
     """
     For the given files, identify all associated test files (i.e. files with the same name, but
     with a "_test.py" suffix).
     """
-    should_include = inclusion_filter(exclude)
 
     test_files = set()
     for file in files:
-        if file.endswith("_test.py"):
+        if file.split("::")[0].endswith("_test.py"):
```
perlinm, line 177: Nice

general_superstaq/check/format_.py
```diff
@@ -25,20 +25,23 @@ def run(
 
     parser.add_argument("--apply", action="store_true", help="Apply changes to files.")
 
-    parsed_args, args_to_pass_black = parser.parse_known_intermixed_args(args)
+    parsed_args, args_to_pass_isort = parser.parse_known_intermixed_args(args)
```
perlinm, line 28: Curious: why this change?

perlinm (commented):
A couple of questions on a quick first pass

general_superstaq/check/format_.py
richrines1: black can't take command line arguments now that we use the `black-configparser` plugin, but isort still does so it seemed more useful to pass them there

perlinm: Ah makes sense

general_superstaq/check/check_utils.py
```diff
@@ -64,45 +64,35 @@ def success(text: str) -> str:
...
+
 
+def exclude_files(files: Iterable[str], exclude: Iterable[str]) -> List[str]:
+    """Returns the files which don't match any of the globs in exclude."""
 
-def inclusion_filter(exclude: Union[str, Iterable[str]]) -> Callable[[str], bool]:
-    """Construct filter that decides whether a file should be included."""
-    if not exclude:
-        return lambda _: True
+    files = sorted(files)
```
perlinm, line 78: Why is this sorted here?

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
...
-            test_file_exists = os.path.isfile(os.path.join(root_dir, test_file))
-            if test_file_exists and should_include(test_file):
+            if os.path.isfile(os.path.join(root_dir, test_file)):
                 test_files.add(test_file)
             elif not silent:
                 print(warning(f"WARNING: no test file found for {file}"))
 
-    return list(test_files)
+    if exclude is not None:
+        exclude = [exclude] if isinstance(exclude, str) else list(exclude)
```
perlinm, line 187: nit: `exclude_files` already accepts an `Iterable`, so this can just be
```suggestion
        exclude = [exclude] if isinstance(exclude, str) else exclude
```

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
     )
 
 
-def get_test_files(*files: str, exclude: Union[str, Iterable[str]] = "", silent: bool) -> List[str]:
+def get_test_files(
+    *files: str, exclude: Union[str, Iterable[str], None] = None, silent: bool
```
perlinm, line 168: nit
```suggestion
    *files: str, exclude: Optional[Union[str, Iterable[str]]] = None, silent: bool
```

general_superstaq/check/check_utils.py
perlinm: The same is true in a few more places in this PR

general_superstaq/check/check_utils.py
```diff
@@ -64,45 +64,35 @@ def success(text: str) -> str:
...
     """
-    match_patterns = [include] if isinstance(include, str) else list(include)
-    matching_files = _check_output("git", "ls-files", *match_patterns).splitlines()
-    should_include = inclusion_filter(exclude)
-    return [file for file in matching_files if should_include(file)]
+    return _check_output("git", "ls-files", "--deduplicate", *include).splitlines()
+
 
+def exclude_files(files: Iterable[str], exclude: Iterable[str]) -> List[str]:
+    """Returns the files which don't match any of the globs in exclude."""
```
perlinm, line 76: Maybe we should make `exclude: Optional[Union[str, Iterable[str]]]` here, rather than converting `exclude` into an `Iterable` in several other places?  Same for `select_files`.

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
...
-        files += get_changed_files(include, exclude, parsed_args.revisions, silent=silent)
-    if not files and search_if_empty:
-        return get_tracked_files(include, exclude)
-    else:
-        return files
+    include = [include] if isinstance(include, str) else list(include)
+    exclude = [exclude] if isinstance(exclude, str) else list(exclude)
+
+    if parsed_args.exclude:
+        exclude += parsed_args.exclude
```
perlinm, line 243: This could be `exclude = list(exclude) + parsed_args.exclude`, if we make `extract_files` accept `exclude: Union[str, Iterable[str]]`

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
...
+
+    files = []
+    globs = []
+    for glob in parsed_args.files:
+        if os.path.isfile(glob.split("::")[0]):  # always include files passed directly as arguments
+            files.append(os.path.relpath(glob, start=root_dir))
+        elif os.path.isdir(glob):  # treat subdirectories the same as the glob "<subdir>/*"
+            globs.append(f"{os.path.relpath(glob, start=root_dir)}/*")
```
perlinm, line 251:
```suggestion
            globs.append(os.path.join(os.path.relpath(glob, start=root_dir), "*"))
```

general_superstaq/check/check_utils.py
```diff
@@ -178,64 +164,105 @@ def _revision_exists(revision: str) -> bool:
...
+        "--force",
+        action="store_true",
+        dest="force_all",
+        help="'Hard force' ~ continue past (i.e. do not exit after) all failing checks.",
+    )
 
     return parser
 
 
 def extract_files(
```
perlinm, line 233: A docstring explaining the logic for what's included vs. excluded would help.  For example (revision needed):
```
Collects a list of files to test, by:
1. Identifying all relevant files tracked by `git` that match `include`,
2. Excluding files according to `exclude`, and finally
3. Keeping only files that match `include`.
Additionally always includes files passed directly as arguments.
```

general_superstaq/check/requirements.py
```diff
@@ -208,14 +205,6 @@ def _inspect_local_version(package: str, latest_version: str) -> None:
         pass
 
 
-def _check_requirements(requirements: List[str]) -> None:
```
perlinm, line 211: good catch

general_superstaq/check/check_utils.py
perlinm: Actually, maybe we should make `exclude: Union[str, Iterable[str]] = ()` everywhere in the check scripts, without the `Optional` part...

general_superstaq/check/pytest_.py
```diff
@@ -36,56 +36,40 @@ def run(
...
-        help="Run pytest on *_integration_test.py files, ignoring dev_tools/*.",
+        help="Run pytest on *_integration_test.py files.",
     )
 
     parser.add_argument("--enable-socket", action="store_true", help="Force-enable socket.")
 
     parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
-    include, exclude = _get_file_search_options(
-        parsed_args.notebook, parsed_args.integration, include, exclude
-    )
```
perlinm, line 47: nice, new version of this is much cleaner :)

general_superstaq/check/pylint_.py
```diff
@@ -27,7 +27,17 @@ def run(
...
+    # perform incremental check by default
+    parser.set_defaults(revisions=[])
+    parser.add_argument(
+        "-a",
+        "--all",
+        action="store_const",
+        const=None,
+        dest="revisions",
+        help="Run pylint on the entire repo.",
+    )
```
perlinm, line 39: The `all` argument appears to be unused at the moment.  Is it still possible to run `pylint` on the entire repo with something like `check/pylint_.py *`?  If so, we can remove the `--all` argument, otherwise we should make this code respect it by setting `parsed.args.revisions = None` if `parsed_args.all == True`.

perlinm (commented):
More comments from a more careful pass

general_superstaq/check/pylint_.py
richrines1: `pylint_.py -a/--all` should still work - setting `action="store_const", const=None, dest="revisions"` here tells the parser to set `revisions=None` iff you pass `-a`/`--all`

general_superstaq/check/check_utils.py
richrines1: i agree, will add

general_superstaq/check/check_utils.py
richrines1: ah i was just trying to avoid filtering out the arguments before passing them to the other scripts, in retrospect i shouldn't have. reverted.

general_superstaq/check/pylint_.py
perlinm: Ohhhhh I see.  Nice :slightly_smiling_face:

general_superstaq/check/check_utils.py
richrines1: ah it's a remnant of debugging. removed. (though changed `get_test_files` slightly so that output remains sorted)

general_superstaq/check/check_utils.py
richrines1: > Actually, maybe we should make `exclude: Union[str, Iterable[str]] = ()` everywhere in the check scripts, without the `Optional` part...

i like it. done.

general_superstaq/check/check_utils.py
richrines1: good call, done.

general_superstaq/check/check_utils.py
richrines1: used slightly different syntax, but done.

general_superstaq/check/check_utils.py
richrines1: done, lmk if you think it's clear enough

general_superstaq/check/check_utils.py
```diff
@@ -63,43 +63,37 @@ def success(text: str) -> str:  # pylint: disable=missing-function-docstring
...
+    """Returns the files which don't match any of the globs in exclude."""
 
-def inclusion_filter(exclude: Union[str, Iterable[str]]) -> Callable[[str], bool]:
-    """Construct filter that decides whether a file should be included."""
-    if not exclude:
-        return lambda _: True
+    exclude = [exclude] if isinstance(exclude, str) else exclude
 
-    exclusions = [exclude] if isinstance(exclude, str) else exclude
+    files = list(files)
```
perlinm, line 79: nit: I think this line doesn't do anything

general_superstaq/check/check_utils.py
```diff
@@ -173,63 +163,112 @@ def _revision_exists(revision: str) -> bool:
...
+
+    Returns:
+        If `parsed_args.files` is empty (i.e. no file paths or globs have been passed to the file
+        parser), a list of tracked files in the active repo branch (relative to `root_dir`) meeting
+        all of the following criteria:
+        1. Path matches at least one of the paths or globs in `include`,
+        2. Path does not match any path or glob in `exclude`,
+        3. Path does not match any path or glob in `parsed_args.exclude` (passed via -x/--exclude),
+        4. If `parsed_args.revisions` is not None, has been modified in the current branch (see
+            documentation for `get_changed_files()` for details).
```
perlinm, line 239: I'm not sure what this is saying

perlinm (commented):
Looks great, just a last bit of minor cleanup.

general_superstaq/check/check_utils.py
richrines1: it's necessary to make the output a list if `exclude` is empty

general_superstaq/check/check_utils.py
```diff
@@ -63,43 +63,38 @@ def success(text: str) -> str:  # pylint: disable=missing-function-docstring
...
-    def should_include(file: str) -> bool:
-        return not any(fnmatch.fnmatch(file, exclusion) for exclusion in exclusions)
+    return files
 
-    return should_include
+
+def select_files(files: Iterable[str], include: Union[str, Iterable[str]]) -> List[str]:
+    """Returns the files which match at least one of the globs in include."""
+
+    files = list(files)
```
richrines1, line 89: realized this was necessary given the `Iterable` return type -- otherwise this function will always return [] if `files` is a generator (because it will be exhausted before the final line)

alternatively we could change the annotation to `files: Collection[str]` to indicate that it shouldn't be a generator... thoughts?

general_superstaq/check/check_utils.py
richrines1: reworded.

general_superstaq/check/check_utils.py
perlinm: I haven't thought through when `Iterable` vs. `Collection` makes sense in cases like this.  I think I may have previously chosen to put `Iterable` everywhere to be "nice" to downstream users, but maybe it's better to be more strict?  Unclear.

Either solution is fine by me.

richrines1: yeah, and we definitely don't have a consistent policy about this. `Iterable` is definitely nicer in that it's as general as possible, but it comes with the risk of weird bugs if we don't consider the possibility of the input being a generator, or clumsier/sometimes less efficient code if we do

in this case the extra `files = list(files)` line doesn't really cost us anything, so i'm leaning toward leaving it and using `Iterable`

perlinm (approved):
LGTM!  Thanks @richrines1!

PR #215, opened 2023-01-25
move gss to subdir

richrines1:
penultimate step in repo merge: moves everything that's unique to general-superstaq into a single subdirectory

related: https://github.com/SupertechLabs/qiskit-superstaq/pull/318, https://github.com/SupertechLabs/cirq-superstaq/pull/442, https://github.com/SupertechLabs/docs-superstaq/pull/40, https://github.com/SupertechLabs/SupermarQ/pull/86

this should be the final pr merged into this repo


.github/workflows/main.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
         env:
           PYPI_API_KEY: ${{ secrets.PYPI_API_KEY }}
         run: |
+          python -m pip install --upgrade pip
```
dowusu-antwi, line 20: Why upgrade `pip`? (If it's what I think, I wonder how we got away with not doing it before...)

dev_tools/update-pypi-version.sh
```diff
@@ -2,6 +2,8 @@
 
 # This publishes general-superstaq to PyPI.
 cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
dowusu-antwi, line 4: We've had this line for a while, but I actually don't know how it works. Doesn't `git rev-parse --show-toplevel` get to the toplevel `gss` directory? If so why do we need to `cd`?

dev_tools/update-pypi-version.sh
```diff
@@ -2,6 +2,8 @@
...
-cd "$(git rev-parse --show-toplevel)"
+toplevel=$(git rev-parse --show-toplevel --path-format=absolute)
+
+cd "$toplevel/general-superstaq"
```
dowusu-antwi, line 7: Cool, so each of `css` and `qss` will do `cd "$toplevel/<client>-superstaq"`. Was the idea to put those in this file, or have a separate one for each? (I.e., publish all vs. publish individually)

dowusu-antwi (commented):
I have a few clarification questions prior to this being review-ready -- hope that's okay!

dev_tools/update-pypi-version.sh
richrines1: though now it occurs to me the line isn't necessary at all after these changes

dev_tools/update-pypi-version.sh
richrines1: my thought was to have them all in the same file, so we can have one `publish all` action

.github/workflows/main.yaml
richrines1: i'm not sure tbh - i just noticed it was done in all the other actions and so figured maybe it should be here too

dev_tools/update-pypi-version.sh
dowusu-antwi: Right, I figured `cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd` would change directories, but then doing the final `cd` on the `pwd` output would seemingly do nothing... But yeah not relevant to repo-merge. Thanks for clarifying!

dev_tools/update-pypi-version.sh
dowusu-antwi: > my thought was to have them all in the same file, so we can have one `publish all` action

Nice, makes sense :+1:

richrines1:
this is ready for review now :)

dowusu-antwi:
I'll review ASAP! (For some reason I don't see it on https://github.com/pulls/review-requested.)

@singular-value we can pair review if you have latitude tomorrow / Monday, but no worries if not.

general-superstaq/.isort.cfg
```diff
@@ -0,0 +1,5 @@
+[settings]
+profile = black
+color_output = True
+line_length = 100
+known_first_party = general_superstaq
```
dowusu-antwi, line 5: This is the same as the one in `css` / `qss`. Is there a reason to not use a single format?

general-superstaq/check
```diff
@@ -0,0 +1 @@
+../check
```
dowusu-antwi, line 1: We're not sure how to read this. Are you creating a new file?

dowusu-antwi (commented):
Looks good, with a couple of comments.

(From Paranav) We might consider creating a new repo to avoid any unforeseen issues (e.g., losing git history - Rich, I think you already considered this?) Otherwise, we should pick the repo with the most stars (`css` v. `qss` v `gss`).

dowusu-antwi:
Looks like SupermarQ has the most stars. (Also, we're definitely merging in that as well right? There's no corresponding PR for it so maybe we postponed...)

general-superstaq/.isort.cfg
dowusu-antwi: Actually, just read the comments [here](https://github.com/SupertechLabs/qiskit-superstaq/pull/318/files#r1109231036) which I think clears it up.

richrines1:
> Also, we're definitely merging in that as well right? There's no corresponding PR for it so maybe we postponed...

yes, but we had been planning to do this in a separate step (given that it has a lot more active users, and especially more that we're not in contact with). So we'd create the monorepo, and once we've seen it's working/stable/etc we'd merge in supermarq. Maybe we should be rethinking this though

> (From Paranav) We might consider creating a new repo to avoid any unforeseen issues (e.g., losing git history - Rich, I think you already considered this?) Otherwise, we should pick the repo with the most stars (css v. qss v gss).

we will definitely be merging the commit histories of every repo into that of the monorepo, so that aspect of the git history will be preserved. We can also transfer all github issues to the monorepo, and obviously anything in the repo we merge into will also be preserved. The one thing that can't be transferred is the github PR history (these will continue to live within the archived repos, which imo seems ok for closed pull requests)

the reason for choosing gss was mostly philosophical: (1) we already use it for "universal" issues, (2) it seems like the natural history to preserve it in-place, in that gss PRs are always relevant to/effectively updates for all the other repos (whereas having our PR history being all-cirq or all-qiskit before a certain date seems weird from a parity perspective). I don't really see any risk in doing so/benefit to creating an entirely new repo (and it's git so everything will be reversible anyway)

I hadn't considered stars at all though -- if it's important @dowusu-antwi is right, we should definitely be merging into supermarq. thoughts? is there any risk in immediately renaming the supermarq repo & merging everything into it?

(sorry long-winded response)

general-superstaq/.isort.cfg
richrines1: yeah it's definitely a bit ugly, hopefully we'll find a better workaround in the future

general-superstaq/check
richrines1: it's a just symlink in the general-superstaq subdirectory to "../check" (so that we can still use `check/xx.py` as before when working within each package directory)

dowusu-antwi: Ah I see now, the directory structure goes from:
```
general_superstaq
|____
     check
```
to
```
general_superstaq -> SuperstaqMonoRepo
|____
     general_superstaq
     |____
          check (symlink)
|____
     check
```

dowusu-antwi:
> we will definitely be merging the commit histories of every repo into that of the monorepo

Awesome! And +1 on closed PRs being in a sense archived -- re-opening PRs seems like a rare event.
 
> is there any risk in immediately renaming the supermarq repo & merging everything into it?

Yeah, leaving to @singular-value (who brought this up) and @vtomole (who was considering stars on SupermarQ recently).

vtomole:
> is there any risk in immediately renaming the supermarq repo & merging everything into it?

The risk i think (Haven't confirmed) is that Github archives links so renaming the SupermarQ repo will lead to the SupermarQ link still working but now linking to the monorepo.

dowusu-antwi (approved):
lgtm!

PR #224, opened 2023-02-14
remove neutral_atom_compile

singular-value:
See also https://github.com/SupertechLabs/cirq-superstaq/pull/439 and https://github.com/SupertechLabs/qiskit-superstaq/pull/315


richrines1 (approved):
lgtm!

PR #225, opened 2023-02-14
require typing-extensions

hallben4:


paaige:
What is this needed for?

richrines1:
@paaige `typing-extensions` is imported in [`gss.typing`](https://github.com/SupertechLabs/general-superstaq/blob/997ce2fa5ff3eb9cc87de9ae99660a80ddd9559a/general_superstaq/typing.py#L3) but we don't include it in our requirements, so `gss` doesn't work when installed in a clean environment. Example:
```
# (after activating a clean virtual env)
pip install general-superstaq
python -c 'import general_superstaq'  # raises ModuleNotFoundError: No module named 'typing_extensions'
```
(this has gotten past our tests because because typing-extensions is required by mypy, and so does get installed by `pip install general-superstaq[dev]`)

paaige (approved):
lgtm!

PR #229, opened 2023-02-16
conditionally require/import typing_extenstions

richrines1:
fixes: https://github.com/SupertechLabs/cirq-superstaq/issues/443

basically, `duet` (which is required by cirq-core) [pins `typing-extensions==3.10.0` for python 3.7](https://github.com/google/duet/blob/f0c396cdf450f56c16b15058f824ab18cf2c4524/requirements.txt#L1). But this version of `typing-extensions` is incompatible with some newer packages (e.g. pylint), so we don't want to require it unless we need to. So this pr limits the requirement/import to python 3.7


general_superstaq/typing.py
```diff
@@ -1,6 +1,11 @@
+import sys
 from typing import Any, Dict, List, Optional, Tuple
 
-from typing_extensions import TypedDict  # compatible with both python 3.7 and 3.8
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
```
paaige, line 7: is this incompatible for < python 3.7

paaige (approved):

general_superstaq/typing.py
richrines1: > do we require at least python 3.7?

yep: https://github.com/SupertechLabs/general-superstaq/blob/8e056285974bd916800dde071eb73df7630f5554/setup.py#L38

richrines1: (currently python 3.7 is the oldest version that still receives security updates, though that will change [in june](https://devguide.python.org/versions/#supported-versions) at which point we should probably revert this and start requiring 3.8)

PR #235, opened 2023-03-02
Add `flake-8-type-checking` to `dev-requirements.txt`

vtomole:
The addition of [this requirement ](https://pypi.org/project/flake8-type-checking/2.3.0/) comes after trying to find ways of automatically doing https://github.com/SupertechLabs/general-superstaq/issues/49. I opted for `from __future__ import annotations` to deal with forward references but if someone fights for wrapping those types around strings instead it's worth discussing.


dowusu-antwi (approved):
everything LGTM

vtomole:
Thanks @dowusu-antwi Will make same PRs in other repos and merge them in order.

general_superstaq/finance.py
```diff
@@ -1,13 +1,17 @@
...
 import numpy.typing as npt
 import qubovert as qv
 
 import general_superstaq as gss
-from general_superstaq import superstaq_client
-from general_superstaq.typing import MaxSharpeJson, MinVolJson
+
+if TYPE_CHECKING:
+    from general_superstaq import superstaq_client
+    from general_superstaq.typing import MaxSharpeJson, MinVolJson
```
richrines1, line 14: instead of adding these lines could we just leave it at `import general_superstaq as gss` and change e.g. `superstaq_client._SuperstaQClient` to `gss.superstaq_client._SuperstaQClient` below?

vtomole: Done.

richrines1: same question re: all the typing imports - would it makes sense to just use `gss.typing.xxx` instead of all the conditional imports?

vtomole: `typing` appears later than most files so this will lead to circular imports. I can make the types public so we can do e.g (`gss.MinVolJson`).

richrines1: ah ok got it. personally i like the public types idea stylistically, but totally fine as is too

richrines1 (approved):
(everything else lgtm!)

general_superstaq/finance.py
vtomole: Making them public seems to give me a circular import on `gss.superstaq_client._SuperstaQClient` in `finance.py`

vtomole: I'll switch them around: make the types public and do explicit type check import on `_SuperstaQCLient`

vtomole:
I've made the types public.

richrines1:
i'm getting this error if i try to run this in a clean python 3.7 env:
```
Traceback (most recent call last):
  File "envs/py37/bin/flake8", line 8, in <module>
    sys.exit(main())
  File "envs/py37/lib/python3.7/site-packages/flake8/main/cli.py", line 22, in main
    app.run(argv)
  File "envs/py37/lib/python3.7/site-packages/flake8/main/application.py", line 363, in run
    self._run(argv)
  File "envs/py37/lib/python3.7/site-packages/flake8/main/application.py", line 350, in _run
    self.initialize(argv)
  File "envs/py37/lib/python3.7/site-packages/flake8/main/application.py", line 330, in initialize
    self.find_plugins(config_finder)
  File "envs/py37/lib/python3.7/site-packages/flake8/main/application.py", line 153, in find_plugins
    self.check_plugins = plugin_manager.Checkers(local_plugins.extension)
  File "envs/py37/lib/python3.7/site-packages/flake8/plugins/manager.py", line 357, in __init__
    self.namespace, local_plugins=local_plugins
  File "envs/py37/lib/python3.7/site-packages/flake8/plugins/manager.py", line 238, in __init__
    self._load_entrypoint_plugins()
  File "envs/py37/lib/python3.7/site-packages/flake8/plugins/manager.py", line 254, in _load_entrypoint_plugins
    eps = importlib_metadata.entry_points().get(self.namespace, ())
AttributeError: 'EntryPoints' object has no attribute 'get'
```
which goes away if i uninstall flake8-type-checking

this appears to be the result of flake8 [needing an outdated version of importlib-metadata](https://github.com/python/importlib_metadata/issues/406) (which later versions of flake8 pin), but upgrading flake8 and/or downgrading importlib-metadata seem to cause a bunch of other errors. i'm not sure why this isn't showing up in the github tests though, so there could well be something weird going on for me locally?

vtomole:
Which versions of `flake-8-type-checking` work for Python 3.7?

dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
 black-configparser~=0.2.0b2  # warning: potential issues https://github.com/danie1k/python-black-configparser/blob/5c6c8d122cdce95e2c7b4ada773cf2f4267c9189/README.md#known-issues
 black[jupyter]~=22.3.0
 flake8~=3.8.4
+flake8-type-checking>=1.0.0
```
richrines1, line 4: wdyt about making this requirement only applicable to python 3.8+ as a fix for https://github.com/SupertechLabs/general-superstaq/pull/235#issuecomment-1453994174
```suggestion
flake8-type-checking>=1.0.0; python_version >= "3.8"
```

vtomole: Sounds good. If anyone in the wild is still running python 3.7 for SuperstaQ dev creates an issue we can view that as a sign to remove `3.7` support. `3.7` is reaching end of life in 3 months anyway.

richrines1:
> Which versions of `flake-8-type-checking` work for Python 3.7?

so far i haven't found any working version

dev-requirements.txt
richrines1: makes sense.

richrines1: > `3.7` is reaching end of life in 3 months anyway.

can't wait :)

richrines1 (approved):
lgtm!

vtomole:
Looks like i need this requirement for downstream PRs so will merge and release.

PR #239, opened 2023-03-08
Update `cq_token` endpoint name for consistency

teaguetomesh:
Opened this PR in response to https://github.com/SupertechLabs/SuperstaQ/pull/2051


vtomole (commented):
Thoughts on https://github.com/SupertechLabs/SuperstaQ/pull/2051/files#r1129936205?

teaguetomesh:
Yeah I think the solution with minimal disruption right now is to make this endpoint `cq_token` to match `ibm_token`. However, maybe we can revise the endpoint naming in the future. About half of the endpoints defined in `superstaq.web.server.py` have names like "verb + noun" (e.g., `update_user_balance`) and the other half are "noun" (e.g., `jobs`). If it becomes an issue we can revisit the naming scheme, and I would probably opt for the more descriptive "verb + noun" style. @vtomole

